PR #402, opened 2014-01-28
CardDAV RFC 5995 - Using POST to Add Members

stevelacey:
I'd be happy to spend some time looking into implementing this if you have any opinion on how to go about doing it, or why it hasn't been done so far? I see it listed as a future feature in the list of related RFCs and implementation statuses on Google Code.

@evert

http://tools.ietf.org/html/rfc5995


DeepDiver1975:
Conflicts are beyond repair -> close

PR #512, opened 2014-09-03
Send a user-agent performing a HTTP request

cbhp:
Some providers block connections with missing user-agents. Also user-agents are useful for analyzing requests. I've added a user-agent that is used in cURL. (ownCloud uses this 1.8 branch)

### DIFF
--- a/lib/Sabre/DAV/Client.php
+++ b/lib/Sabre/DAV/Client.php
@@ -314,7 +314,8 @@ class Client {
             CURLOPT_RETURNTRANSFER => true,
             // Return headers as part of the response
             CURLOPT_HEADER => true,
-
+            // Send identification
+            CURLOPT_USERAGENT => 'SabreDAV DAV client',
             // For security we cast this to a string. If somehow an array could
             // be passed here, it would be possible for an attacker to use @ to
             // post local files.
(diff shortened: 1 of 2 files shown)


lib/Sabre/DAV/Client.php
```diff
@@ -314,7 +314,8 @@ public function request($method, $url = '', $body = null, $headers = array()) {
             CURLOPT_RETURNTRANSFER => true,
             // Return headers as part of the response
             CURLOPT_HEADER => true,
-
+            // Send identification
+            CURLOPT_USERAGENT => 'SabreDAV DAV client',
```
Hywan: I guess `SabreDAV client` is enough.

cbhp: I just have taken the name from sabre-dav/lib/DAV/Client.php:8 * SabreDAV DAV client

Hywan: Ok fine for me :-).

Hywan:
:+1:

Asking a peer review from @evert :-).

Hywan:
@cbhp Oh yes, please, rebase your patch into a single commit (because the first one fails). Thanks!

DominikTo:
Would it make sense to use something like `'SabreDAV Client' . '/' . DAV\Version::VERSION´?
Could be handy for debugging.

staabm:
> Would it make sense to use something like 'SabreDAV Client' . '/' . DAV\Version::VERSION?

it would also be more in line with what actual browsers do... :+1:

Hywan:
What having a `DAV\Client::AGENT_NAME` constant also?

evert:
Those things all sound pretty good :)

evert:
The main reason I didn't immediately merge was sort of for every reason mentioned in this ticket so far.. I wanted to clean it up a bit, and also make sure that this was done correctly for branches later than 1.8. 

1.8 is not _that_ important anymore.. 2 version ago at this point, but it may still be nice to set the user agent and spam access logs worldwide ;)

PR #521, opened 2014-09-15
PostgreSQL tests

mattes:
I started to add tests for PostgreSQL and already fixed some issues. Though, some tests still fail. From my point of view, it is not safe to use PostgreSQL at the moment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,15 @@ matrix:
 services:
   - mysql
 
+addons:
+  postgresql: "9.3"
+
 before_script:
   - mysql -e 'create database sabredav'
 #  - composer self-update
   - composer install --prefer-source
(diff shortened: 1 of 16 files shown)


fabianwolf:
I just tried to get these changes working with the newest version of sabre/dav.
1) The lastInsertId did fix some issues,
but i was still stuck with foreign key violation errors.
As far as i can see these "foreign key on delete cascades" are not needed, because sabre/dav cleans up the database itself, these are also not part of the mysql schema, so easiest thing would be removing them from postgres schema.
2) I just copied the intval fix as it worked with the tests, but i dont know if this gives expected behaviour everywhere.
3) Postgres indeed seems to return streams, so adjusting the tests to accept streams is a good idea.

evert:
My goal is now to get this ready for version 3.2. I might not use this PR as a basis though, as it's quite old now.

evert:
Pull request #825 will succeed this.

PR #564, opened 2014-11-21
ETag is based on the file's mtime. #performance

Hywan:
Quoting the documentation of the `DAV\IFile::getETag` method:

> An ETag is a unique identifier representing the current version of the file. If the file changes, the ETag MUST change.

In [`DAV\FSExt\File`](https://github.com/fruux/sabre-dav/blob/93e61954fd82dfc397055cec5a10a636cf982a45/lib/DAV/FSExt/File.php#L115), we use the [`md5_file` function (documentation)](http://php.net/md5_file). I was thinking: I wonder if it can cause performance issue? I think yes because the [`md5_file` function (implementation)](http://lxr.php.net/xref/PHP_5_6/ext/hash/hash_md.c#135) reads the entire file in order to compute the MD5 hash. So with a file of 5Gb, it can hurt. I was then thinking: why not having a better way to compute the Etag of a file? Well, the [`mtime`](http://php.net/filemtime)! In order to protect the data behind the Etag, we encode this data with a SHA1.

However, I have also noticed that `FS\File` returns a `null` Etag while we can apply the same approach. So this PR fixes these two guys + respective tests.

Finally, a last patch ensures that the `fstat`'s cache is clear before computing the `ETag`. We optimize this cache invalidation by clearing only the cache of the current file, not for all the FS.

### DIFF
--- a/lib/DAV/FS/Directory.php
+++ b/lib/DAV/FS/Directory.php
@@ -39,7 +39,8 @@ class Directory extends Node implements DAV\ICollection, DAV\IQuota {
     function createFile($name, $data = null) {
 
         $newPath = $this->path . '/' . $name;
-        file_put_contents($newPath,$data);
+        file_put_contents($newPath, $data);
+        clearstatcache(true, $newPath);
 
     }
 
@@ -53,6 +54,7 @@ class Directory extends Node implements DAV\ICollection, DAV\IQuota {
 
         $newPath = $this->path . '/' . $name;
         mkdir($newPath);
+        clearstatcache(true, $newPath);
 
     }
 
--- a/lib/DAV/FS/File.php
+++ b/lib/DAV/FS/File.php
@@ -70,7 +70,11 @@ class File extends Node implements DAV\IFile {
      */
     function getETag() {
 
-        return '"' . sha1(filemtime($this->path)). '"';
+        return '"' . sha1(
+            fileinode($this->path) .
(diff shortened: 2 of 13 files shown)


Hywan:
ping?

evert:
Apache uses a combination of the INode number, filemtime and size combined. This seems like a smart approach.

evert:
https://developer.yahoo.com/performance/rules.html#etags says:

> The ETag format for Apache 1.3 and 2.x is inode-size-timestamp. Although a given file may reside in the same directory across multiple servers, and have the same file size, permissions, timestamp, etc., its inode is different from one server to the next. 

So that leaves size + filemtime =) so

Hywan:
(`rebase master`)

Hywan:
Ready for a review :-].

untitaker:
Apparently PHP caches ``stat`` results: http://php.net/manual/en/function.clearstatcache.php

evert:
@untitaker is right clearstatcache() should be called before returning the etag _after_ any update (so createFile/put).

staabm:
`clearstatcache` should be given the path to the file, so it won't clear the whole cache, but just the parts we are working on.

Hywan:
My version included `clearstatcache`. I will update my PR right now.

Hywan:
I rebase on master.

Hywan:
\o/

PR #570, opened 2014-11-26
Introduce the FSIterator.

Hywan:
First step consisted to remove `scandir` and replace it by `FilesystemIterator` (see 8c81bc3). However, while it is a little bit faster and cross-FS, it does not solve the problem of memory consumption (and inherited CPU consumption) when having a lot of children.

Thus, this second step consists to solve this bottleneck. Instead of:

  1. iterating 1 time on children,
  2. transform them into nodes and
  3. return all the nodes as an array, which will be iterated a 2nd time,

we build an iterator that:

  1. given a collection (`ICollection`) and a path,
  2. generate and return a filesystem iterator that transforms children into nodes on demand.

Children are iterated later.

Bonus: a filter can be added, useful in `FSExt/Directory.php` for instance.

### DIFF
--- a/lib/DAV/FS/Directory.php
+++ b/lib/DAV/FS/Directory.php
@@ -85,24 +85,13 @@ class Directory extends Node implements DAV\ICollection, DAV\IQuota {
     }
 
     /**
-     * Returns an array with all the child nodes
+     * Returns an iterator with all the child nodes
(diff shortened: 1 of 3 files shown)


evert:
Closing in favor of #681

Issue #572, opened 2014-11-26
PHP warning on header sent with HEAD


Hywan:
Hello :-),

Here is a very basic server in order to reproduce the issue:

``` php
<?php
// index.php

require 'vendor/autoload.php';
use Sabre\DAV;

$rootDirectory = new DAV\FS\Directory('public');
$server = new DAV\Server($rootDirectory);

$server->exec();
```

in `public/`, 2 files. One containing `foobarbaz`, another one of 1Gb, created with:

``` sh
$ dd if=/dev/zero of=1Gb bs=1024 count=1048576
```

Start the server:

``` php
$ php -S 127.0.0.1:1234 -t .
```

And cURL the first simple little file:

``` sh
$ curl -X HEAD 127.0.0.1:1234/Foo --verbose
```

everything is fine.
Now, cURL the big file of 1Gb:

``` sh
$ curl -X HEAD 127.0.0.1:1234/1Gb --verbose
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 1234 (#0)
> HEAD /1Gb HTTP/1.1
> User-Agent: curl/7.38.0
> Host: 127.0.0.1:1234
> Accept: */*
>
< HTTP/1.1 200 OK
< Host: 127.0.0.1:1234
< Connection: close
< X-Powered-By: PHP/5.6.4-dev
< X-Sabre-Version: 2.2.0-alpha1
< Last-Modified: Wed, 26 Nov 2014 15:58:32 GMT
< Content-Type: application/octet-stream
< Content-Length: 1073741824
<
* transfer closed with 1073741824 bytes remaining to read
* Closing connection 0
```

The `Content-Length` is wrong. In addition, on the server side, we have:

```
[Wed Nov 26 17:10:40 2014] 127.0.0.1:52042 [200]: /Foo
[Wed Nov 26 17:11:39 2014] PHP Warning:  Cannot modify header information - headers already sent by (output started at …/vendor/sabre/http/lib/Sapi.php:70) in …/vendor/sabre/http/lib/Sapi.php on line 58
[Wed Nov 26 17:11:39 2014] PHP Warning:  Cannot modify header information - headers already sent by (output started at …/vendor/sabre/http/lib/Sapi.php:70) in …/vendor/sabre/http/lib/Sapi.php on line 63
[...]

PR #577, opened 2014-12-01
Fix for #572

evert:


RealRancor:
@evert As a follow-up of https://github.com/owncloud/android/issues/1074 and https://github.com/owncloud/core/issues/13786:

Seems that fix should have been included in 2.1.2:

https://github.com/fruux/sabre-dav/blob/3.1.1/CHANGELOG.md#212-2014-12-10

but never made it into that release.

AnrDaemon:
I'd echo the same request. We just [got hit by this very issue](http://forum.ubuntu.ru/index.php?topic=271478.msg2168758#msg2168758) and the proposed change indeed resolved the problem the user was having.

evert:
I'm fairly sure that that change got reverted later, because it was a BC break.

AnrDaemon:
Does that mean that mobile clients are useless right now?

evert:
I'm not sure I understand your question. This ticket refers to a very specific issue. It got reverted later, but a new solution _was_ added on the 2.1 branch. The more correct fix got added in 3.0, but afaik this should not be a problem in stock sabre/dav on the 2.1 branch.

If this is causing a new problem, please just open a new ticket with a reproducible test case. "It causes an error on owncloud" is not enough information for me to figure out what's wrong. If it's an issue in owncloud, open a ticket on that project, not here.

AnrDaemon:
The issue is still present in ownCloud 8.2.3, the solution from this PR still works.

evert:
Then you should report the issue at Owncloud. I'm locking this thread now.

PR #588, opened 2015-01-05
Remove an unused variable

### LINKED ISSUE #586: SCHEDULE-STATUS should not contain an explanatory text
I was looking at the `IMipPlugin` sabredav provides and was wondering about the `scheduleStatus` values being set in the `Itip\Message`. Reading through http://tools.ietf.org/html/rfc6638#section-7.3 it looks like there is no room for explanatory text and it should instead be a list of statcodes.

I haven't checked what scheduling emails actually look like so I may be wrong here, but I think its a bug.

Another trivial issue you might want to just piggyback here, `$server` is defined as a class variable, but the `initialize` function does not set it.

Hywan:
Thanks @kewisch for the notice!

Partly fix #586.


evert:
Thank you :)

PR #592, opened 2015-01-07
Add support of whitelists of authorized paths

Hywan:
The goal is to introduce a whitelist of authorized paths per auth backend. Why? Because we don't systematically want to check if the user is authenticated on all paths. For instance, a path like `/signin` or `/signup` are not good candidate for authentifications since this is the user entry point to authenticate ;-). Same issue happens with OAuth, `/access_token` and `/authorize`.

Thoughts?

### DIFF
--- a/lib/DAV/Auth/Backend/AbstractBasic.php
+++ b/lib/DAV/Auth/Backend/AbstractBasic.php
@@ -167,8 +167,8 @@ abstract class AbstractBasic implements BackendInterface {
      * Example of authorized paths:
      *
      * [
-     *     '/signup',
-     *     '/signin'
+     *     'signup',
+     *     'signin'
      * ]
      *
      * @param RequestInterface $request
--- a/lib/DAV/Auth/Backend/AbstractBasic.php
+++ b/lib/DAV/Auth/Backend/AbstractBasic.php
@@ -33,13 +33,6 @@ abstract class AbstractBasic implements BackendInterface {
     protected $realm = 'sabre/dav';
 
     /**
-     * Whitelist of authorized paths.
-     *
-     * @var array
-     */
-    protected $whiteList = [];
-
-    /**
      * This is the prefix that will be used to generate principal urls.
      *
      * @var string
@@ -67,18 +60,6 @@ abstract class AbstractBasic implements BackendInterface {
     function setRealm($realm) {
 
         $this->realm = $realm;
-
-    }
-
-    /**
-     * Sets the whitelist of authorized paths.
-     *
-     * @param array $paths
-     * @return void
-     */
-    function setWhiteList(array $paths) {
-
-        $this->whiteList = $paths;
 
     }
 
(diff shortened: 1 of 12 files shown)


lib/DAV/Auth/Plugin.php
```diff
@@ -142,8 +142,15 @@ function beforeMethod(RequestInterface $request, ResponseInterface $response) {
         if (!$this->backends) {
             throw new \Sabre\DAV\Exception('No authentication backends were configured on this server.');
         }
+
         $reasons = [];
-        foreach($this->backends as $backend) {
+        $backends = $this->backends;
+        foreach($backends as $i => $backend) {
+
+            if (in_array($request->getPath(), $this->getWhiteList())) {
```
staabm: `getWhitelist` could be called once, pre-loop

staabm: Or should it be $backend->gerWhitelist() ?

Hywan: It's `$backend->…` yes ;-). I am writing tests right now and it is fixed.

lib/DAV/Auth/Backend/AbstractBasic.php
```diff
@@ -142,5 +161,23 @@ function challenge(RequestInterface $request, ResponseInterface $response) {
...
+     *
+     * [
+     *     'signup',
+     *     'signin'
+     * ]
+     *
+     * @param RequestInterface $request
+     * @return array
+     */
+    function getWhiteList(RequestInterface $request) {
```
evert: Curious why the argument is needed.

lib/DAV/Auth/Backend/AbstractBasic.php
```diff
@@ -33,6 +33,13 @@
     protected $realm = 'sabre/dav';
 
     /**
+     * Whitelist of authorized paths.
+     *
+     * @var array
+     */
+    protected $whiteList = [];
```
evert: I'm wondering if we could just move the whitelist to the plugin, instead of all the backends. This would make this feature a lot simpler.

I can definitely see that the contents of the whitelist is

evert:
This is my non-blocking comment:

In the future my plan is to to allow anything in the system to tell the authentication plugin to *not* require authentication for certain requests. This would not be whitelist based, but rather rule-based.

Specifically, I want to get to a point where users can use the ACL system to allow certain operations to be done by unauthenticated users.

I'm pointing this out, because when we finally have that feature, the whitelist feature would get deprecated.

lib/DAV/Auth/Backend/AbstractBasic.php
Hywan: If the whitelist depends on a variable somewhere else. Just in case.

lib/DAV/Auth/Backend/AbstractBasic.php
Hywan: So, you suggest to move the whitelists into the plugin instead of the backend? I am going to do that but without any rebases.

Hywan:
Is it then necessary to go further with this one (even if it's done)? Introducing a feature we know we are going to deprecated soon does not sound a good idea for our users :-/.

evert:
@Hywan well this was my way of leaving the decision with you =)

Hywan:
I don't close this issue as a reminder. But it's stopped.

evert:
Closing in favor of #657

PR #595, opened 2015-01-14
Handle floatingTimes in ICSExportPlugin

armin-hackmann:
A calendars timezone is already considered for events with floating times in e.g. FreeBusy requests (see #555) but not when expanding events in the ICSExportPlugin. This Pull Request adds that.


evert:
Looks great!

PR #598, opened 2015-01-26
Normalizing URIs before comparing them.

### LINKED ISSUE #597: URI equivalence is not good enough
Trying to setup the 'caldav-tester' test-suite, I've found that I was unable to even really start the tests.

The problem lies in the fact that the caldav tester automatically encodes a url such as:

    /~evert/sabre/dav/

into

    /%7Eevert/sabre/dav/

While this may not be 100% the preferred URL, sabredav should still properly deal with this situation.

evert:
This fixes issues when clients incorrectly encode parts of the path, such as
the tilde.

Fixes #597


composer.json
```diff
@@ -18,6 +18,7 @@
         "sabre/vobject": "~3.3.4",
         "sabre/event" : "~2.0.0",
         "sabre/http" : "~3.0.4",
+        "sabre/uri" : "dev-normalize as 0.0.2",
```
Hywan: Maybe we should wait `dev-normalize` to be merged into `master` of sabre/uri?

evert: Yea I will change it back after merging that. I needed a working sabredav branch for testing

Issue #581, opened 2014-12-06
scheduling : issue with local delivery


ma2thieu:
I can't get local delivery to work with scheduling, if the 2 users (attendant and organizer) are on the same server the invitation is still sent by email.

When I look at `$iTipMessage->recipient` in [scheduleLocalDelivery](https://github.com/fruux/sabre-dav/blob/master/lib/CalDAV/Schedule/Plugin.php#L366) there is a `mailto:` in front of the email address and the function returns because it can't find the principal.

evert:
Fixed in #604

PR #604, opened 2015-02-09
Issue #581 : Fixed blank principalPrefix in findByUri

soydeedo:
The principalPrefix sent to searchPrincipals was set to a literal blank
string and would never return a valid result unless no prefix was used.

https://github.com/fruux/sabre-dav/issues/581


evert:
Great patch!

soydeedo:
Thanks for the quick response!

Hywan:
:+1:

PR #606, opened 2015-02-10
Use VObject\Writer in CalDAV and CardDAV exporters

Hywan:
When sabre/dav will use a version of sabre/vobject that includes `Sabre\VObject\Writer`, we could merge this PR.


Hywan:
Is it still relevant?

DeepDiver1975:
> Is it still relevant?

I think not -> close

PR #610, opened 2015-02-23
Prevent enumeration of absolute path using Browser Plugin

LukasReschke:
Previously URIs like `?sabreAction=asset&assetName=../../../../../../../../../../../Users/lreschke/Programming/core/3rdparty/sabre/dav/lib/DAV/Browser/assets/sabredav.css` could have been used as  `realpath` does resolve them properly.

As discussed in https://github.com/owncloud/core/commit/1edd6d7d0b15a3fd0ad7b20107ac0c603f63204c#commitcomment-9804349

@evert As discussed.

### DIFF
--- a/lib/DAV/Browser/Plugin.php
+++ b/lib/DAV/Browser/Plugin.php
@@ -456,6 +456,7 @@ HTML;
      *
      * @param string $assetName
      * @return string
+     * @throws DAV\Exception\NotFound
      */
     protected function getLocalAssetPath($assetName) {
 
@@ -463,6 +464,10 @@ HTML;
         $path = $assetDir . $assetName;
 
         // Making sure people aren't trying to escape from the base path.
+        $path = str_replace('\\', '/', $path);
+        if (strpos($path, '/../') !== FALSE || strrchr($path, '/') === '/..') {
+            throw new DAV\Exception\NotFound('Path does not exist, or escaping from the base path was detected');
+        }
         if (strpos(realpath($path), realpath($assetDir)) === 0 && file_exists($path)) {
             return $path;
         }
--- a/lib/DAV/Browser/Plugin.php
+++ b/lib/DAV/Browser/Plugin.php
@@ -465,7 +465,7 @@ HTML;
 
         // Making sure people aren't trying to escape from the base path.
         $path = str_replace('\\', '/', $path);
-        if (strpos($path, '/../') !== FALSE || strrchr($path, '/') === '/..') {
+        if (strpos($path, '/../') !== false || strrchr($path, '/') === '/..') {
             throw new DAV\Exception\NotFound('Path does not exist, or escaping from the base path was detected');
         }
         if (strpos(realpath($path), realpath($assetDir)) === 0 && file_exists($path)) {


LukasReschke:
```
FILE: /home/travis/build/fruux/sabre-dav/lib/DAV/Browser/Plugin.php
--------------------------------------------------------------------------------
FOUND 1 ERROR(S) AFFECTING 1 LINE(S)
--------------------------------------------------------------------------------
 468 | ERROR | TRUE, FALSE and NULL must be lowercase; expected "false" but
     |       | found "FALSE"
--------------------------------------------------------------------------------
UPGRADE TO PHP_CODESNIFFER 2.0 TO FIX ERRORS AUTOMATICALLY
--------------------------------------------------------------------------------
```

Let me find out what black magic that is :see_no_evil:

evert:
Writing a unittest for this one

LukasReschke:
> Writing a unittest for this one

Need to find out how to start the SabreDAV test suite on my Mac. It failed always hard with:

```
PHP Warning:  include(/Users/lreschke/Programming/core/3rdparty/sabre/dav/tests/../vendor/autoload.php): failed to open stream: No such file or directory in /Users/lreschke/Programming/core/3rdparty/sabre/dav/tests/bootstrap.php on line 9
PHP Stack trace:
PHP   1. {main}() /usr/local/Cellar/phpunit/4.3.4/libexec/phpunit-4.3.4.phar:0
PHP   2. PHPUnit_TextUI_Command::main() /usr/local/Cellar/phpunit/4.3.4/libexec/phpunit-4.3.4.phar:605
PHP   3. PHPUnit_TextUI_Command->run() phar:///usr/local/Cellar/phpunit/4.3.4/libexec/phpunit-4.3.4.phar/phpunit/TextUI/Command.php:138
PHP   4. PHPUnit_TextUI_Command->handleArguments() phar:///usr/local/Cellar/phpunit/4.3.4/libexec/phpunit-4.3.4.phar/phpunit/TextUI/Command.php:148
PHP   5. PHPUnit_TextUI_Command->handleBootstrap() phar:///usr/local/Cellar/phpunit/4.3.4/libexec/phpunit-4.3.4.phar/phpunit/TextUI/Command.php:651
PHP   6. PHPUnit_Util_Fileloader::checkAndLoad() phar:///usr/local/Cellar/phpunit/4.3.4/libexec/phpunit-4.3.4.phar/phpunit/TextUI/Command.php:817
PHP   7. PHPUnit_Util_Fileloader::load() phar:///usr/local/Cellar/phpunit/4.3.4/libexec/phpunit-4.3.4.phar/phpunit/Util/Fileloader.php:77
PHP   8. include_once() phar:///usr/local/Cellar/phpunit/4.3.4/libexec/phpunit-4.3.4.phar/phpunit/Util/Fileloader.php:93
```

etc.

Any suggestions? :smile:

lib/DAV/Browser/Plugin.php
```diff
@@ -463,6 +464,10 @@ protected function getLocalAssetPath($assetName) {
         $path = $assetDir . $assetName;
 
         // Making sure people aren't trying to escape from the base path.
+        $path = str_replace('\\', '/', $path);
+        if (strpos($path, '/../') !== false || strrchr($path, '/') === '/..') {
```
staabm: path might also be home-relative (starting with `~`)

evert:
The problem is that the unittests expect sabre/dav to be loaded as the main package, not as a dependency.

So the easiest is to just do a separate clone of sabredav, or you could run `composer install` in the sabre/dav directory (but that will also create another `vendor` directory there).

Alternatively you could also create your own `bootstrap.php` that uses the owncloud autoloader instead of PHP's.

lib/DAV/Browser/Plugin.php
evert: This is true and makes me wonder two things:

1. The vulnerability @LukasReschke wanted to fix is the ability to discover paths on the system outside of the sabredav installation path. As far as I know this is only a windows concern, because on posix the path will be expanded without checking for existance of the paths. Is this correct?
2. If this is not correct, then `~` is vulnerable as well.. and perhaps there are others. Is there a more robust OWASP-approved way to do this?

LukasReschke: > path might also be home-relative (starting with ~)

$path is prepended with $assetDir and thus the path may never start with `~`

evert: Ah... `~` doesn't work in the middle of a path? I suppose that makes sense.

LukasReschke: > The vulnerability @LukasReschke wanted to fix is the ability to discover paths on the system outside of the sabredav installation path. As far as I know this is only a windows concern, because on posix the path will be expanded without checking for existance of the paths. Is this correct?

Just for clarification purposes: Actually my only problem here was that this allows to enumerate the path where SabreDAV is installed (also on OS X and pretty sure also Linux). 

It was however not possible to access arbitrary files on the OS or enumerate them. – That prevented this patch already pretty good since we have it added in 2013.

(so this is not really a very important fix and only nice-to-have as an additional hardening)

evert: Fair enough then!

LukasReschke: > Ah... ~ doesn't work in the middle of a path? I suppose that makes sense.

That is correct from what I can tell. Also defined as such at https://www.gnu.org/software/libc/manual/html_node/Tilde-Expansion.html

> It’s a standard part of shell syntax that you can use ‘~’ at the beginning of a file name to stand for your own home directory. You can use ‘~user’ to stand for user’s home directory.
> [...]
> Tilde expansion applies to the ‘~’ plus all following characters up to whitespace or a slash. **It takes place only at the beginning of a word**, and only if none of the characters to be transformed is quoted in any way.

<hr/>

That said if we want to harden this further the best way is always a whitelist. So in this case it might make sense to use a regex to limit the possible path. Though I don't know what would all be allowed. Possibly also subdirectories?

So something like `([a-z]+)/([a-z]+)\.(css|png)` – though I'm VERY bad at regex ;-)

PR #611, opened 2015-02-23
Escape some more values

LukasReschke:
While they should not contain user-supplied input at all it's always good to sanitize more data in case somehow the API got misused by developers.

As discussed in https://github.com/owncloud/core/commit/1edd6d7d0b15a3fd0ad7b20107ac0c603f63204c#commitcomment-9804349

@evert


evert:
Thank you. Gonna get this on 2.1 as well.

PR #613, opened 2015-02-24
[Client] Use file not found exception on 404 HTTP status in propFind

patkar:
Use file not found exception on 404 HTTP status in propFind, because a 404 is a exceptional case. So the calling code/library can catch on the 404 and handle and ignore all other exceptions, which must be handled in the application on a higher level or be shown.

Be best would be an exception chooser based, which chooses the correct exception based on the HTTP status, but at the moment I had time for this fix only.

### DIFF
--- a/lib/DAV/Client.php
+++ b/lib/DAV/Client.php
@@ -240,6 +240,10 @@ class Client extends HTTP\Client {
         $response = $this->send($request);
 
         if ((int)$response->getStatus() >= 400) {
+            if ((int)$response->getStatus() === 404) {
+                throw new Exception\NotFound($url . ' not found');
+            }
+
             throw new Exception('HTTP error: ' . $response->getStatus());
         }
 
--- a/tests/Sabre/DAV/ClientTest.php
+++ b/tests/Sabre/DAV/ClientTest.php
@@ -163,6 +163,20 @@ XML;
 
     }
 
+    /**
+     * @expectedException \Sabre\DAV\Exception\NotFound
+     */
+    function testPropFindNotFound() {
+
+        $client = new ClientMock([
+            'baseUri' => '/',
+        ]);
+
+        $client->response = new Response(404, []);
+        $client->propfind('foo', ['{DAV:}displayname', '{urn:zim}gir']);
+
+    }
+
     function testPropFindDepth1() {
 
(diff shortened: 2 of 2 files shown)


evert:
The reverse of this change has been made some time ago. Instead of using the NotFound exception, this needs to emit a Sabre\Http\ClientHttpException, but this is a breaking change and can only be done for version 2.2 (master branch).

patkar:
But without this findProp exceptions are silly, because we cannot check which HTTP code emitted the exception.. What about a new `ClientException` which extends the base `\Sabre\Dav\Exception` and encapsulates the returned HTTP status code or the whole response, so the consuming library (e.g. `league/flysystem`) can check if it's a 404 and catch it or rethrow?

Btw. BC breaks should bump the major version, not the minor. BC breaks would cause a v3.0 (please use semver.org).

patkar:
Oh, and btw. your change sometimes ago, throwing the base exception instead of a not found exception was a BC break.

evert:
If you set `$throwExceptions` to true in the client, you _do_ get exceptions with specific status codes, _and_ you get access to the full http response.

As for using semver, we've decided against it. The second digit in the version number is for BC breaking changes. This is well documented, and I rather not have this discussion here and stay on-topic.

> Oh, and btw. your change sometimes ago, throwing the base exception instead of a not found exception was a BC break.

Correct, but this change happened in a BC-breaking version, was announced and documented, and didn't happen in a minor revision.

patkar:
Oh this is by default false :O So we should set this to true and cat a `ClientHttpException`, were we can check the status code?

Then there is only these annoying lines in `propFind()` which doesn't mind the `$throwExceptions`.. Should it be removed or what?

evert:
@patkar Indeed, I think setting it to true is all you need.

For the long turn:

1. In 2.2, we'll just turn that Sabre\DAV\Exception to a ClientHttpException as well. That was an oversight, which sucks.
2. Turning throwExceptions on by default is not a bad idea, but I actually think the better move is to remove it altogether and have 1 consistent behavior (which is to always throw exceptions). I think that that would be a lot more obvious. This can also only be done in a BC breaking version (of both sabre/http and sabre/dav), so I think we should just do that there =)

PR #616, opened 2015-03-03
Use an iterator to get properties for nodes on demand

icewind1991:
This saves having to keep the properties of all nodes in memory before we generate the xml.

Additionally since it now re-uses the same `PropFind` object to store the result it saves having to create and gc additional objects for each file in the propfind.

When testing using ownClouds's sabre implementation this saves 50% of memory when doing a propfind on a folder containing 10k files ([comparison](https://blackfire.io/profiles/compare/b3efda1d-741a-4f26-9e8e-9aeda345a6ce/graph))

### DIFF
--- a/lib/CardDAV/VCFExportPlugin.php
+++ b/lib/CardDAV/VCFExportPlugin.php
@@ -87,7 +87,7 @@ class VCFExportPlugin extends DAV\ServerPlugin {
      * @param array $nodes
      * @return string
      */
-    function generateVCF(array $nodes) {
+    function generateVCF($nodes) {
 
         $output = "";
 
--- a/lib/DAV/PropFindIterator.php
+++ b/lib/DAV/PropFindIterator.php
@@ -1,0 +1,106 @@
+<?php
+
+namespace Sabre\DAV;
+
+class PropFindIterator implements \Iterator, \ArrayAccess
+{
+    /**
+     * @var PropFind
+     */
+    private $propFind;
+
+    /**
+     * @var \SeekableIterator
+     */
+    private $nodes;
+
+    /**
+     * @var Server
+     */
+    private $server;
+
+    public function __construct(PropFind $propFind, \SeekableIterator $nodes, Server $server)
+    {
+        $this->propFind = $propFind;
+        $this->nodes = $nodes;
+        $this->server = $server;
+    }
+
+    public function current()
+    {
+        /**
+         * @var INode $node
+         * @var string $path;
+         */
+        list($node, $path) = $this->nodes->current();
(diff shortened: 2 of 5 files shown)


evert:
This is definitely the approach I want to take, but I want to do it using generators instead. The resulting code will be a lot more elegant. So if you are willing to prepare a patch that uses generators instead of iterators, that might definitely help this move forward.

evert:
But if you're not interested, I understand as well. For one, generators require PHP 5.5. I just don't think this intermediate step is worth it.

icewind1991:
Do you have a planned timescale for droping 5.4 support?

evert:
2.2 will support 5.4, 2.3 will require 5.5. Hard to say exactly when each is released, we don't really have a fixed release schedule.

lib/CardDAV/VCFExportPlugin.php
```diff
@@ -87,7 +87,7 @@ function httpGet(RequestInterface $request, ResponseInterface $response) {
      * @param array $nodes
      * @return string
      */
-    function generateVCF(array $nodes) {
```
Hywan: We should type with `Traversable`.

lib/DAV/PropFindIterator.php
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    public function valid()
+    {
+        return $this->nodes->valid();
+    }
+
+    public function offsetSet($offset, $value)
+    {
+        throw new Exception();
```
Hywan: A message should be at least provided :-).

lib/DAV/Server.php
```diff
@@ -1632,11 +1604,11 @@ function getResourceTypeForNode(INode $node) {
...
      * If 'strip404s' is set to true, all 404 responses will be removed.
      *
-     * @param array $fileProperties The list with nodes
-     * @param bool strip404s
+     * @param array | \Iterator $fileProperties The list with nodes
+     * @param bool $strip404s
      * @return string
      */
-    function generateMultiStatus(array $fileProperties, $strip404s = false) {
+    function generateMultiStatus($fileProperties, $strip404s = false) {
```
Hywan: We should type with `Traversable`.

evert:
Note my other messages @Hywan , I don't think this is something we should accept as a feature. I'm glad the PR was made early in the process to not waste a lot of time.

Hywan:
@evert I read your messages after having comment the patch :-p. But this is not bad to have a feedback even if the PR is not yet accepted :-).

evert:
Closing this. This should be solved by using PHP 5.5 generators. See #681

PR #617, opened 2015-03-11
RecurrenceIterator -> Recur\EventIterator

massar:
Avoid using 

https://github.com/fruux/sabre-vobject/wiki/Sabre-VObject-RecurrenceIterator

which is deprecated, and does not work anymore in the current tree


evert:
Thank  you!

PR #620, opened 2015-04-01
Retrieve property information for files

evert:
This PR improves the browser plugin by allowing users to get properties for files.

Todo:

* [x] Merge #602 
* [x] Rebase this PR
* [ ] Unittest the new code.

### DIFF
--- a/lib/DAV/Browser/Plugin.php
+++ b/lib/DAV/Browser/Plugin.php
@@ -62,9 +62,27 @@ class Plugin extends DAV\ServerPlugin {
     function initialize(DAV\Server $server) {
 
         $this->server = $server;
+        $this->server->on('method:GET', [$this,'httpGetEarly'], 90);
         $this->server->on('method:GET', [$this,'httpGet'], 200);
         $this->server->on('onHTMLActionsPanel', [$this, 'htmlActionsPanel'],200);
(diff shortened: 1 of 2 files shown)


lib/DAV/Browser/Plugin.php
```diff
@@ -60,12 +60,30 @@ function __construct($enablePost=true) {
...
+     *
+     * @param RequestInterface $request
+     * @param ResponseInterface $response
+     * @return bool
+     */
+    function httpGetEarly(RequestInterface $request, ResponseInterface $response) {
+
+        $params = $request->getQueryParameters();
+        if (isset($params['sabreAction']) && $params['sabreAction']==='info') {
+            return $this->httpGet($request,$response);
```
staabm: will this method then be invoked twice? 

2nd time on line https://github.com/fruux/sabre-dav/pull/620/files#diff-932be52ff072a432a7d1d44b2822f58dR64 then?

evert: The way the `method:` events work, is that as soon as there's an event handler that is willing to handle the request, the remaining events will no longer be executed.

If the event handler (so `httpGetEarly`) returns `false`, then event handling stops and the second one will never happen.

PR #623, opened 2015-04-01
Remove the need of temp files from the handling of Range requests

dratini0:
This commit requires fruux/sabre-http#45 to work.

The other day, I tried to play a 10 gig HD movie from my Owncloud server through GVFS. It failed, and hanged my server by forcing PHP to create humongous temp files. (see owncloud/core#15321) As it turns out, the problem was with HTTP Range requests: SabreDAV decided to create temp files with the the entire contents of the range (GVFS requested ranges like 40960-). So I went ahead to create this pair of small patches.

I am looking forward to hearing your opinion on the matter. Sorry for not opening a discussion beforehand, but discussing it would have taken longer than writing it.

And yes, with these applied, I *have* managed to play the video, at least in a separate SabreDAV install.

### DIFF
--- a/lib/DAV/CorePlugin.php
+++ b/lib/DAV/CorePlugin.php
@@ -170,14 +170,16 @@ class CorePlugin extends ServerPlugin {
 
             }
 
-            // fseek will return 0 only if $stream is seekable (and -1 otherwise)
             // for a seekable $body stream we simply set the pointer
             // for a non-seekable $body stream we read and discard just the
-            //  right amount of data
-            if ((fseek($body, $start, SEEK_SET)) !== 0) {
-                $consume_block = 4096;
+            // right amount of data
+            if (stream_get_meta_data($body)['seekable']) {
+                fseek($body, $start, SEEK_SET);
+            } else {
+                $consumeBlock = 8192;
                 for($consumed = 0; $start - $consumed > 0; ){
-                    $consumed += strlen(fread($body, min($start - $consumed, $consume_block)));
+                    if(feof($body)) throw new Exception\RequestedRangeNotSatisfiable('The start offset (' . $start . ') exceeded the size of the entity (' . $consumed . ')');
+                    $consumed += strlen(fread($body, min($start - $consumed, $consumeBlock)));
                 }
             }
 
--- a/lib/DAV/CorePlugin.php
+++ b/lib/DAV/CorePlugin.php
@@ -170,24 +170,21 @@ class CorePlugin extends ServerPlugin {
 
             }
 
-            // New read/write stream
-            $newStream = fopen('php://temp','r+');
-
-            // fseek will return 0 only if $streem is seekable (and -1 otherwise)
-            // for a seekable $body stream we set the pointer write before copying it
-            // for a non-seekable $body stream we set the pointer on the copy
-            if ((fseek($body, $start, SEEK_SET)) === 0) {
-                stream_copy_to_stream($body, $newStream, $end - $start + 1, $start);
-                rewind($newStream);
-            } else {
-                stream_copy_to_stream($body, $newStream, $end + 1);
-                fseek($newStream,$start, SEEK_SET);
+            // fseek will return 0 only if $stream is seekable (and -1 otherwise)
+            // for a seekable $body stream we simply set the pointer
+            // for a non-seekable $body stream we read and discard just the
+            //  right amount of data
+            if ((fseek($body, $start, SEEK_SET)) !== 0) {
+                $consume_block = 4096;
+                for($consumed = 0; $start - $consumed > 0; ){
+                    $consumed += strlen(fread($body, min($start - $consumed, $consume_block)));
+                }
             }
 
             $response->setHeader('Content-Length', $end - $start + 1);
             $response->setHeader('Content-Range','bytes ' . $start . '-' . $end . '/' . $nodeSize);
             $response->setStatus(206);
-            $response->setBody($newStream);
+            $response->setBody($body);
 
         } else {
 
--- a/tests/Sabre/DAV/ServerRangeTest.php
+++ b/tests/Sabre/DAV/ServerRangeTest.php
@@ -38,7 +38,7 @@ class ServerRangeTest extends AbstractServer{
          );
 
         $this->assertEquals(206, $this->response->status);
-        $this->assertEquals('st c', stream_get_contents($this->response->body));
+        $this->assertEquals('st c', stream_get_contents($this->response->body, 4));
 
     }
 
@@ -70,7 +70,7 @@ class ServerRangeTest extends AbstractServer{
          );
 
         $this->assertEquals(206, $this->response->status);
-        $this->assertEquals('st contents', stream_get_contents($this->response->body));
+        $this->assertEquals('st contents', stream_get_contents($this->response->body, 11));
 
     }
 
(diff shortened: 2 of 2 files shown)


lib/DAV/CorePlugin.php
```diff
@@ -170,24 +170,21 @@ function httpGet(RequestInterface $request, ResponseInterface $response) {
...
-            if ((fseek($body, $start, SEEK_SET)) === 0) {
-                stream_copy_to_stream($body, $newStream, $end - $start + 1, $start);
-                rewind($newStream);
-            } else {
-                stream_copy_to_stream($body, $newStream, $end + 1);
-                fseek($newStream,$start, SEEK_SET);
+            // fseek will return 0 only if $stream is seekable (and -1 otherwise)
+            // for a seekable $body stream we simply set the pointer
+            // for a non-seekable $body stream we read and discard just the
+            //  right amount of data
```
Hywan: One extra space before `right`.

lib/DAV/CorePlugin.php
```diff
@@ -170,24 +170,21 @@ function httpGet(RequestInterface $request, ResponseInterface $response) {
...
+            if ((fseek($body, $start, SEEK_SET)) !== 0) {
+                $consume_block = 4096;
```
Hywan: Please, use `$consumeBlock` instead of `$consumeBlock`.

lib/DAV/CorePlugin.php
```diff
@@ -170,24 +170,21 @@ function httpGet(RequestInterface $request, ResponseInterface $response) {
...
+                for($consumed = 0; $start - $consumed > 0; ){
+                    $consumed += strlen(fread($body, min($start - $consumed, $consume_block)));
+                }
```
Hywan: The algorithm is correct.

lib/DAV/CorePlugin.php
```diff
@@ -170,24 +170,21 @@ function httpGet(RequestInterface $request, ResponseInterface $response) {
...
-                stream_copy_to_stream($body, $newStream, $end - $start + 1, $start);
-                rewind($newStream);
-            } else {
-                stream_copy_to_stream($body, $newStream, $end + 1);
-                fseek($newStream,$start, SEEK_SET);
+            // fseek will return 0 only if $stream is seekable (and -1 otherwise)
+            // for a seekable $body stream we simply set the pointer
+            // for a non-seekable $body stream we read and discard just the
+            //  right amount of data
+            if ((fseek($body, $start, SEEK_SET)) !== 0) {
```
Hywan: Maybe this is a good opportunity to use a proper way to check seekability of the stream.
```php
$seekable = stream_get_meta_data($body)['seekable'];
```
It avoids to handle an error internally. In any way this is faster.

lib/DAV/CorePlugin.php
```diff
@@ -170,24 +170,21 @@ function httpGet(RequestInterface $request, ResponseInterface $response) {
...
+                $consume_block = 4096;
+                for($consumed = 0; $start - $consumed > 0; ){
+                    $consumed += strlen(fread($body, min($start - $consumed, $consume_block)));
```
Hywan: Instead of counting the result from `fread` and summing it to `$consumed`, we can use the following (faster) code:
```php
fread($body, min($start - $consumed, $consume_block));
$consumed = ftell($body);
```
If the stream is not seekable, we can still get the position of the pointer with `ftell`. See the following program:
```php
<?php

$descriptors = [
    0 => ['pipe', 'r'],
    1 => ['pipe', 'w']
];

$processus = proc_open('ls', $descriptors, $pipes);

var_dump(stream_get_meta_data($pipes[1]));
var_dump(ftell($pipes[1]));

$data = stream_get_contents($pipes[1]);
var_dump($data);
var_dump(ftell($pipes[1]));
```
What the meta data says: The `stream_type` of `$pipes[1]` is `STDIO` and this is a non-seekable stream.
The first `ftell` returns 0.
We consume some data.
The last `ftell` returns a number greater than 0.

So it works on non-seekable stream.

Hywan: Can you confirm this is faster (if you have the opportunity to test)? I think it is.

dratini0: I can, I am working on it right now.

dratini0: Hmm. Strangely enough, my method seems faster, even if it involves counting characters. I was testing using /dev/zero and the following code:

```php
<?php
$start = 1<<30;
$consume_block = 4096;
if($argv[1] != 1){
    echo 0 . "\n";
    $startTime = microtime(true);
        for($consumed = 0; $start - $consumed > 0; ){
            $consumed += strlen(fread(STDIN, min($start - $consumed, $consume_block)));
        }
    echo microtime(true)-$startTime . "\n";
} else {
    echo 1 . "\n";
    $startTime = microtime(true);
        for($consumed = 0; $start - $consumed > 0; ){
            fread(STDIN, min($start - $consumed, $consume_block));
            $consumed = ftell(STDIN);
        }
    echo microtime(true)-$startTime . "\n";
}
```

For the record: PHP 5.6.7-1 (cli)

Hywan:
Good idea and I like the new solution. More clever and straightforward.

However (I think it is not related to this new patch but it applies in all situations), what if we ask for a range from 42-100 and the stream only contains 7 bytes. Another scenario is if the stream is corrupted. The `$consumed` variable will never reach the `$start` variable and we will have an infinite loop. I think we must detect if `$consumed` stagnates (does not increase) during 2 steps in the loop, **or** if we reach the end of the stream. The latter solution is more elegant because we can use `feof`. Nevertheless, remind the stream can be corrupted, the algorithm could therefore never reach the end of the stream and `feof` is inefficient. Consequently, I propose to add an invariant checking `$consumed` increases at each step, something like this:
```php
for ($previousConsumed = $consumed = 0; $start - $consumed > 0;) {

    fread($body, min($start - $consumed, $consume_block));
    $consumed = ftell($body);

    if (!($previousConsumed < $consumed)) {
        // break
        throw new Exception(…); // don't know which exception
    }

    $previousConsumed = $consumed;
}
```
We could add the invariant `$previousConsumed < $consumed` in the `for(…; $invariant;)` but we would not be able to throw an exception. And I think it clarifies the code.

lib/DAV/CorePlugin.php
dratini0: For method 0, I was getting 1.4187359809875 seconds to 1.716873884201, and for method 1, I was getting 1.6698949337006 to 1.7286839485168

Hywan: Hmm, this is not a surprise in fact. See http://lxr.php.net/xref/PHP_TRUNK/Zend/zend_types.h#499:
```c
#define Z_STRLEN(zval)              Z_STR(zval)->len
```
The length of the string is stored in the `zval` structure. However, `ftell` does not have such a cache. I was thinking that the returned string from `fread` was not stored in memory because it is not assigned…

Could you check the memory please (more difficult, do something trivial)?

dratini0: Hmm. If you mean memory use, after increasing $consumeBlock to 2^26, I have managed to run both with with ```ulimit -v 303000``` and niether with ```ulimit -v 329000```. If you mean debugging the PHP executable itself, sorry, I wouldn't really do that.

evert:
Love the idea and the quality of the patch. Really great.

dratini0:
That is nice to hear! :smile:

PR #635, opened 2015-04-16
`s/httpPropPatch/httpProppatch`

Hywan:
The method did not have the correct case.

I assign @evert for the merge (don't know if we should merge on `master`, I guess it would).


DominikTo:
Maybe we should change it in https://github.com/Hywan/sabre-dav/blob/httpproppatch/lib/DAV/CorePlugin.php#L41 instead. In other places `propPatch` seems to be used. Apparently also for events that are fired etc.

evert:
Hmm.. yea I've been using casing such as propFind, propPatch, mkCol. Are we calling it `httpProppatch` somewhere?

Hywan:
So I should rename `httpPropfind` to `httpPropFind` too. Will do that after some sleep ;-).

Hywan:
Done.

evert:
thank you =)

PR #637, opened 2015-04-20
Set $propertyList otherwise it is not initialized

massar:
Otherwise one gets a mysterious HTTP 500 that is not logged.

Only way to figure out what goes wrong is to use:
https://github.com/fruux/sabre-dav/issues/397

This one liner fixes this.

### DIFF
--- a/lib/CalDAV/Plugin.php
+++ b/lib/CalDAV/Plugin.php
@@ -420,7 +420,8 @@ class Plugin extends DAV\ServerPlugin {
 
         $needsJson = $report->contentType === 'application/calendar+json';
 
-        $timeZones = [];
+	$timeZones = [];
+	$propertyList = [];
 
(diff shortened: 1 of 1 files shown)


lib/CalDAV/Plugin.php
```diff
@@ -420,7 +420,8 @@ function calendarMultiGetReport($report) {
 
         $needsJson = $report->contentType === 'application/calendar+json';
 
-        $timeZones = [];
+	$timeZones = [];
```
staabm: please fix indent

massar: It is a one line patch.....

evert: The difference between submitting a patch that matches our CS versus not doing that, is that I could have just hit the merge button and be done with it. But it's no big deal, I just made the same change in 6dd9ced

PR #642, opened 2015-04-27
Remove PHP5.2 code

Hywan:
Noticing this PHP5.2 code. Now we dropped it, we can clean the code.

### DIFF
--- a/lib/DAVACL/Plugin.php
+++ b/lib/DAVACL/Plugin.php
@@ -416,47 +416,44 @@ class Plugin extends DAV\ServerPlugin {
 
         $privs = $this->getSupportedPrivilegeSet($node);
 
-        $flat = [];
-        $this->getFPSTraverse($privs, null, $flat);
+        $fpsTraverse = null;
+        $fpsTraverse = function ($priv, $concrete, &$flat) use (&$fpsTraverse) {
 
-        return $flat;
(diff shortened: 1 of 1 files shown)


lib/DAVACL/Plugin.php
```diff
@@ -416,47 +416,44 @@ static function getDefaultSupportedPrivilegeSet() {
...
+
+                }
 
             }
 
-        }
+        };
+
+        $flat = [];
+        $fpmsTraverse($privs, null, $flat);
```
staabm: should this be `$fpsTraverse`?

evert:
Awesome =) this was indeed quite old school.

Hywan:
I used an old trick for the recursivity. Don't know if you have something better :-/.

evert:
That's also how i normally do that, You could use a y-combinator but it's overkill :P

Hywan:
How?

evert:
It's mostly a joke, I think this is the best way =)

evert:
https://php100.wordpress.com/2009/04/13/php-y-combinator/

Hywan:
Oh, this one: https://en.wikipedia.org/wiki/Fixed-point_combinator#Y_combinator, I though http://www.ycombinator.com/ ;-).

PR #643, opened 2015-04-28
Automated coding standards fixing.

evert:
Note that this should not be merged until the migration document for sabredav 2.1 -> 2.2 is written. I usually do that by reading manually through the diff, and merging this beforehand would make that a lot more work.


lib/CalDAV/Xml/Request/CalendarMultiGetReport.php
```diff
@@ -113,7 +113,7 @@ static function xmlDeserialize(Reader $reader) {
         }
 
         $obj = new self();
-        foreach($newProps as $key=>$value) {
+        foreach($newProps as $key => $value) {
```
Hywan: Why the missing space after `foreach` is not added?

lib/DAV/Exception.php
```diff
@@ -9,7 +9,6 @@
  * @author Evert Pot (http://evertpot.com/)
  * @license http://sabre.io/license/ Modified BSD License
  */
-
```
Hywan: Sure?

lib/DAV/FS/Directory.php
```diff
@@ -76,7 +77,7 @@ function getChild($name) {
 
         if (is_dir($path)) {
 
-            return new Directory($path);
+            return new self($path);
```
Hywan: In this case, we should even simplify by `return new static($path);`, but maybe in another PR?

lib/DAV/FSExt/Directory.php
```diff
@@ -85,11 +85,11 @@ function getChild($name) {
         $path = $this->path . '/' . $name;
 
         if (!file_exists($path)) throw new DAV\Exception\NotFound('File could not be located');
-        if ($name=='.' || $name=='..') throw new DAV\Exception\Forbidden('Permission denied to . and ..');
+        if ($name == '.' || $name == '..') throw new DAV\Exception\Forbidden('Permission denied to . and ..');
 
         if (is_dir($path)) {
 
-            return new Directory($path);
+            return new self($path);
```
Hywan: Same here

lib/DAV/PartialUpdate/Plugin.php
```diff
@@ -166,8 +165,8 @@ function httpPatch(RequestInterface $request, ResponseInterface $response) {
 
         $this->server->emit('afterWriteContent', [$path, $node]);
 
-        $response->setHeader('Content-Length','0');
-        if ($etag) $response->setHeader('ETag',$etag);
+        $response->setHeader('Content-Length', '0');
+        if ($etag) $response->setHeader('ETag', $etag);
```
Hywan: Hm, we definitively need a “brace fixer” for this kind of cases.

lib/DAV/Server.php
```diff
@@ -446,7 +445,7 @@ function invokeMethod(RequestInterface $request, ResponseInterface $response, $s
         if (!$this->emit('beforeMethod:' . $method, [$request, $response])) return;
         if (!$this->emit('beforeMethod', [$request, $response])) return;
 
-        if (Server::$exposeVersion) {
+        if (self::$exposeVersion) {
```
Hywan: Yup, maybe `static::` would be great here too.

Hywan:
Sounds good for me. However, we should also apply this the `tests/` folder, thoughts?

lib/CalDAV/Xml/Request/CalendarMultiGetReport.php
evert: Removed that per a conversation with dominik about that.

lib/DAV/Exception.php
evert: Yea this was one of the symfony rules that I think makes sense for us too. No space between the docblock and the thing that it's documenting.

lib/DAV/FS/Directory.php
evert: One reason why somebody might subclass this is to for example change what `getName()` returns. This was actually reasonably common to me, I needed a different directory name in the tree compared to the actual path on disk.

Had I used `return new static` here, it meant that any sub-collection would also have that hardcoded name.

I think by default we should _never_ do `return new static` because not only does it make assumptions about how a class is subclassed, it also makes assumptions about the constructor signature.

The constructor is exempt from the liskov principle, so I think in general this is quite a bad idea.

lib/DAV/Server.php
evert: The reason this was a static variable in the first place, was because it had to be accessed by unrelated functions.

Turning this into `static` from `self` would give people the incorrect illusion that they can subclass the server class and override the default there. This would not work. So therefore I think it's best to just keep it as `self::` as well.

Hywan:
Check https://github.com/fruux/sabre-cs/issues/14.

PR #645, opened 2015-04-28
Created a per-user collection collection.

evert:
This patch adds a new feature, that's basically a `/home` for sabre/dav.

I know we're thinking of adding something like this to sabre/katana. I think this is all we need, but not sure.

Needs unittests.


Hywan:
How to use it?

I guess in sabre/katana we must add quota, MIME type, public directory and encryption. I will sum this up everywhere. I was planning to work on this today, glad you started it. Any idea when this PR will be merged?

evert:
The only thing missing here is unittests =) after that I can immediately merge it.

PR #648, opened 2015-04-30
Supporting the callable function

nheimann1:
The current implementation had the following problem

    Fatal error: Call to undefined function \ProductName\CardDAV\BasicAuthCallback::validateUserPass() 
    in ../SabreDAV/vendor/sabre/dav/lib/DAV/Auth/Backend/BasicCallBack.php on line 56

At least while working with the following PHP version: 

    PHP 5.5.9-1ubuntu4.5 (cli) (built: Oct 29 2014 11:59:10) 
    Copyright (c) 1997-2014 The PHP Group
    Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.3, Copyright (c) 1999-2014, by Zend Technologies

The fix was made according to [PHP Documentation :: Callable](http://php.net/manual/en/language.types.callable.php)

### DIFF
--- a/lib/DAV/Auth/Backend/BasicCallBack.php
+++ b/lib/DAV/Auth/Backend/BasicCallBack.php
@@ -53,7 +53,7 @@ class BasicCallBack extends AbstractBasic {
     protected function validateUserPass($username, $password) {
 
         $cb = $this->callBack;
-        return $cb($username, $password);
+        return call_user_func($cb, $username, $password);
 
     }
 


evert:
Hey!

I think there's some confusion going on, because this works:


```php
<?php

class Hello {

    function test() {
        echo "hp\n";
    }

}

$hello = new Hello();
$cb = [$hello, 'test'];
$cb();
```

Could you post a more complete example of where this is breaking?

nheimann1:
Hey,
The problem starts when trying to work with static functions. Please look on the code below.
 
I think that the advantage with my fix is, that you can choose the right syntax according the situation, the previous code limits you to write only in one syntax, which makes perfect sense when working with non static methods. but when working with static methods it enforce you to create an instance of the class, which kind of break the purpose of static methods.

```
<?php

class CallableClass
{
	public static function staticMethod()
	{
		echo 'static method';
	}

    public function nonStaticMethod()
    {
		echo 'non static method';
    }
}


class InvokingClass
{
	public function invoke1($callBack)
	{
		$callBack(); 
	}
	
	public function invoke2($callBack)
	{
		call_user_func($callBack);
	}

}

$invokingClass = new InvokingClass();
$callableClass = new CallableClass();


//<code>//<result> [ - <my comment>]

//Static methods
//$invokingClass->invoke1('CallableClass::staticMethod');//Fatal error: Call to undefined function CallableClass::staticMethod()
//$invokingClass->invoke2('CallableClass::staticMethod');//static method - For static methods, I think that this syntax is clearer

//$invokingClass->invoke1([$callableClass, 'staticMethod']);//static method - For static methods, I think that creating an instance is not a good practice 
//$invokingClass->invoke2([$callableClass, 'staticMethod']);//static method - For static methods, I think that creating an instance is not a good practice 



//Non static methods
//$invokingClass->invoke1('CallableClass::nonStaticMethod');//Fatal error: Call to undefined function CallableClass::nonStaticMethod() - make sense since this is a non static method and have no instace of the class
//$invokingClass->invoke2('CallableClass::nonStaticMethod');//non static method - It calls the default constructor, eventhough I think that this kind of use is a bad practice 

//$invokingClass->invoke1([$callableClass, 'nonStaticMethod']);// non static method
//$invokingClass->invoke2([$callableClass, 'nonStaticMethod']);//non static method

```

evert:
What about this syntax:

    $invokingClass->invoke1(['CallableClass', 'staticMethod']);

I have to say though that I'm shocked that the callable typehint permits  `class::method` syntax but then it cant be called without call_user_func. I think this should be considered a php bug.

I want to do a bit more investigation here.

nheimann1:
Your proposed syntax is working well `$invokingClass->invoke1(['CallableClass', 'staticMethod']);` and it has more sense that making an instance of the class.

I agree with you that something looks wrong with the documentation and/or the implementation of PHP

I understand your  desire to investigate deeper here. 

All I can say is that using the `call_user_func($callBack)` will allow the user to choose his own syntax rather than forcing him to use only one of them. For me for example it took some time before I understood what was going on, since I tried this `class::method` all the time and nothing worked, and only after ruling out all the other possible problems (there were none) I looked deeper into the callable documentation, than some more trial and error got me to understand the problem.

evert:
Blogged it here:

http://evertpot.com/on-callables-and-closures/

I'm definitely on the side of 'it should work' and 'it's a php bug', especially since there's a workaround, I'm not sure if I want to make this change.

I know it's just a one-liner here, but it's a bit more than that. It doesn't make sense to just change it there. If we consider this is a necessity it means we need to do it absolutely everywhere and make it part of our coding standards.

nheimann1:
Your project, your decision, since both decisions have pros and cons, and none of them is absolute superior compare to the other.

But I would suggest documenting it really good if you choose to keep the current implementation. I would document it both in code (in the description of the method) and also here http://sabre.io/dav/authentication/

evert:
@nheimann1 it's treated as a bug and now fixed in php master: https://github.com/php/php-src/pull/1264#issuecomment-101248927

It will take a while before you can take advantage, but at least it will be working correctly one day =) For now you indeed need to work around it, but perhaps it feels slightly less bad now you know you are working around a PHP bug, and not a sabre/dav bug.

nheimann1:
@evert Thanks for the update.
I already implemented the workaround in my code, and at the end I'm not using a static function, so anyway it is better to use the supported syntax in that case.

I think that referencing the code to this discussion is advised in order for other people not to waste time since they are using the "wrong" syntax.

PR #650, opened 2015-05-11
Fix calendar deletion

ErrOrnAmE:
Deleting calendar previously didn't deleted every occurrence of the calendar in the Calendar Changes table, but only the one with the id corresponding to the calendarId (even with no link between them)


evert:
Shit!

Thanks about this. This is a big miss.

evert:
I'll backport this to all relevant earlier versions as well.

PR #653, opened 2015-05-19
`getChild` works with a class-child of `Directory`

Hywan:
We simply replace `self` by `static`.

Needed for sabre/katana.


evert:
As discussed on slack, I find this to be bad design.

PR #660, opened 2015-05-20
Principal could break html output for non-href principals.

evert:


lib/DAVACL/Xml/Property/Principal.php
```diff
@@ -123,6 +124,36 @@ function xmlSerialize(Writer $writer) {
...
+        switch ($this->type) {
+
+            case self::UNAUTHENTICATED :
+                return '<em>unauthenticated</em>';
+            case self::AUTHENTICATED :
+                return '<em>authenticated</em>';
+            case self::HREF :
+                return parent::toHtml($html);
+            case self::ALL :
+                return '<em>all</em>';
```
staabm: should have a defined `default` case

PR #667, opened 2015-05-26
FIX for webdav.mediencenter.t-online.de

OliverKohlDSc:
https://webdav.mediencenter.t-online.de returns invalid response code.

e.g.
{"reqId":"f9a1c394b98108e4e5ca62bf47829c64","remoteAddr":"81.189.45.224","app":"PHP","message":"Undefined offset: 2 at \/var\/www\/owncloud\/3rdparty\/sabre\/dav\/lib\/Sabre\/DAV\/Client.php#569","level":3,"time":"2015-03-25T18:25:48+00:00","method":"GET","url":"\/index.php\/apps\/files\/ajax\/getstoragestats.php?dir=External%2FT-Cloud%2FTests"}

as documented here: https://github.com/owncloud/3rdparty/pull/169

### DIFF
--- a/lib/Sabre/DAV/Client.php
+++ b/lib/Sabre/DAV/Client.php
@@ -557,8 +557,15 @@ class Client {
 
                 $propStat->registerXPathNamespace('d', 'urn:DAV');
                 $status = $propStat->xpath('d:status');
-                list($httpVersion, $statusCode, $message) = explode(' ', (string)$status[0],3);
-
+                
+                // This resolves issues that are flooding the log file due to some malformed
+                // WebDAV response e.g. T-Mobile's cloud named "Mediencenter"
+                // This has the positve side-effect of a huge speed inporovement on erros
+                if (count($status) < 1 || empty($status[0]) || count($status) < 3)
+                    $statusCode='200';
+                else
+                    list($httpVersion, $statusCode, $message) = explode(' ', (string)$status[0],3);
+                
                 // Only using the propertymap for results with status 200.
                 $propertyMap = $statusCode==='200' ? $this->propertyMap : array();
 


lib/Sabre/DAV/Client.php
```diff
@@ -557,8 +557,15 @@ public function parseMultiStatus($body) {
 
                 $propStat->registerXPathNamespace('d', 'urn:DAV');
                 $status = $propStat->xpath('d:status');
-                list($httpVersion, $statusCode, $message) = explode(' ', (string)$status[0],3);
-
+                
+                // This resolves issues that are flooding the log file due to some malformed
+                // WebDAV response e.g. T-Mobile's cloud named "Mediencenter"
+                // This has the positve side-effect of a huge speed inporovement on erros
+                if (count($status) < 1 || empty($status[0]) || count($status) < 3)
```
staabm: why do you check the count of `$status`  twice?

why should this change boost performance?
(I think it might be slow in your case, because of extensive error-logging?)

evert:
If we're interested in supporting broken behavior from other servers, we'll need a _lot_ more detail about what the server is reponding, what it should be responding, and why this fix would work. A unittest would also be required.

This patch might work for your specific case, but it actually turns every status code in a 200, but I assume that that's just a bug in your patch.

Breaking standards compliance is not something we take lightly, you'll need to come up with some proper motivation and show that you've attempted to log the bug first at the other vendor's system.

lib/Sabre/DAV/Client.php
OliverKohlDSc: Before an exception has been thrown per each loop pass and logged in owncloud. This fix sets the status code immediatly to 200 instead of reporting an exception each time an element has been found. Think of that procedure for 23000 items. The timeout occurs faster than time is needed for the execution of the code.

OliverKohlDSc:
I agree with the unit test. But it's actually desired to return 200 in every case. It's not a bug. Why? If you already have the list of items $propStat, why would you then return any other http. The item is already there and it can be retrieved, it just lacks some xml markup.

evert:
It's not... the error code gives you information about the property itself, including whether a property is available or not. If there's a 404 in there it means that the server tells us 'the resource exists, but the property does not'. This is critical behavior in webdav.

If an exception is thrown and an error is logged for every iteration of that loop, perhaps that's where your problem is.

OliverKohlDSc:
@evert: You are totally right.

PR #668, opened 2015-05-29
provide information about vcf plugin

mgee:

PR #669, opened 2015-05-29
add button to export an addressbook in the browser

mgee:


evert:
The broken unittest here is related to #668, not to this change.

evert:
Messed up the issue numbers :/ I think this is fine to merge otherwise @Hywan , not that concerned about a unittest for this thing.

mgee:
@evert thanks for merging, sorry didn't notice the unittest...

staabm:
looks good

PR #675, opened 2015-06-16
Client add support for NTLM auth

k42b3:
This pull request adds an AUTH_NTLM constant to the DAV client which can be used when talking 
to an windows webdav server. Or is there any specific reason why only basic and digest auth 
is supported at the moment?


evert:
Hi!

There's no specific reason other than: nobody tested it. Does this patch work for your use-case?

k42b3:
Hi, thx for the fast response. Yes with this fix I could connect to an Windows WebDAV share using NTLM (Windows Authentication).

PR #686, opened 2015-07-14
searchPrincipals with email: case insentive, and no wildcards

vmire:
Schedule plugin does not process local delivery if attendee's email has a different character case than corresponding Principal in the database. (ex: test@domain.com vs Test@domain.com).

On the other hand, there should not be any wildcard in the search query in order to avoid wrong results : 
ex: 
test@domain.com search criteria must not have those results:
another-test@domain.com
test@domain.com.uk


lib/DAVACL/PrincipalBackend/PDO.php
```diff
@@ -252,8 +252,8 @@ function searchPrincipals($prefixPath, array $searchProperties, $test = 'allof')
                     $values[] = '%' . $value . '%';
                     break;
                 case '{http://sabredav.org/ns}email-address' :
-                    $query .= ' AND email LIKE ?';
-                    $values[] = '%' . $value . '%';
+                    $query .= ' AND lcase(email) = lcase(?)';
```
Hywan: Is `lcase` an SQL standard? This is supposed to work with SQLite and PostgreSQL too.

vmire: LCASE seems to work with MySQL and PGSQL but not SQLite.

LOWER is a synonym which works with all of them

Hywan: So, yup, rebase your patch please :-].

vmire: done

evert:
Hi @vmire , the current behavior for that method is to a certain extend deliberate. It's also invokes when a user searches for principals (for autocomplete).

We only later on also overloaded the method to search principals for delivering messages, but this is indeed wrong :/

What we need to do is update `findByUri` instead and remove false positives after calling searchPrincipals. 

Reference: https://github.com/fruux/sabre-dav/blob/master/lib/DAVACL/PrincipalBackend/AbstractBackend.php#L35

If you're up for this change, let me know... otherwise I will close this PR and re-open it as a separate ticket.

vmire:
I understand.

I do not really like this kind of compensatory measure, but I’m ok.

I would rather add a new parameter to searchPrincipals in order to specify explicitly the wildcard behavior (not clear in function description).
	searchPrincipals($prefixPath, array $searchProperties, $test = ’allof’, $wildcardSearch = TRUE)
However, it should introduce a backward compatibility issue on BackendInterface.

evert:
@vmire , I would agree... the correct solution is to just implement findByUri() correctly in Sabre\DAVACL\PrincipalBackend\PDO as it's intended.

The existing behavior to fall back on searchPrincipals was only ever intended as a fallback, and to make it easy for people to upgrade versions when `findByUri()` was introduced.

So one possibility is that:

1. `findByUri` gets correctly implemented in the `PDO` class.
2. `findByUri` in `AbstractBackend` _still_ filters out false positives
3. In the future we remove `findByUri` from `AbstractBackend`.

I only asked for number 2, but if you feel like doing number 1 as well then I think that will also satisfy your earlier concern. Adding new parameters is not really an option, especially when we want to maintain backwards compatibility.

evert:
Hey @vmire ,

The patch mostly looks good, but you should do a CS cleanup. If the dev tools were installed, you should just be able to run: `./bin/sabre-cs-fixer fix lib/`

lib/DAVACL/PrincipalBackend/PDO.php
```diff
@@ -280,6 +280,55 @@ function searchPrincipals($prefixPath, array $searchProperties, $test = 'allof')
...
+     * @return string
+     */
+	function findByUri($uri, $principalPrefix) {
+		$idx = strpos($uri,":");
+		if($idx===FALSE){
+			return null;
+		}
+		
+		$scheme = substr($uri,0,$idx);
+		$value = substr($uri,$idx+1);
```
evert: Replace with:

list($scheme, $value) = explode($uri, ':', 2);

vmire:
I don’t have dev tools installed.
How can I install it?

lib/DAVACL/PrincipalBackend/PDO.php
```diff
@@ -240,26 +240,26 @@ function updatePrincipal($path, DAV\PropPatch $propPatch) {
...
      */
     function searchPrincipals($prefixPath, array $searchProperties, $test = 'allof') {
+		if(count($searchProperties) == 0) return [];	//No criteria
 
-        $query = 'SELECT uri FROM ' . $this->tableName . ' WHERE 1=1 ';
+        $query = 'SELECT uri FROM ' . $this->tableName . ' WHERE ';
         $values = [];
         foreach ($searchProperties as $property => $value) {
-
+			$column;
```
Hywan: What is this line ;-)?

lib/DAVACL/PrincipalBackend/PDO.php
```diff
@@ -280,6 +280,55 @@ function searchPrincipals($prefixPath, array $searchProperties, $test = 'allof')
...
+				$query = 'SELECT uri FROM ' . $this->tableName . ' WHERE lower(email)=lower(?)';
+				$stmt = $this->pdo->prepare($query);
+		        $stmt->execute([ $value ]);
+	        
+	        	while($row = $stmt->fetch(\PDO::FETCH_ASSOC)) {
+	           		// Checking if the principal is in the prefix
+	           		list($rowPrefix) = URLUtil::splitPath($row['uri']);
+					if($rowPrefix !== $principalPrefix) continue;
+					
+	            	$uri = $row['uri'];
```
Hywan: You have some issue with tabs vs. spaces I suppose :-)?

Hywan:
@vmire `composer install` would be enough to get them.

evert:
Probably `composer install --dev` actually, but could be wrong.

Hywan:
@evert `--dev` is run by default. You have to use `--no-dev` to avoid requiring `require-dev` packages.

evert:
I think that's true for update, not install. But it looks like @vmire already figured it ou

vmire:
Yes, composer install works.

vmire:
Are there still an issue that prevents the merge?

evert:
Yes, we need to make sure that this new stuff is 100% unittested =)

lib/DAVACL/PrincipalBackend/PDO.php
```diff
@@ -281,6 +280,52 @@ function searchPrincipals($prefixPath, array $searchProperties, $test = 'allof')
...
+        $uri = null;
+        switch ($scheme){
+            case "mailto":
+                $query = 'SELECT uri FROM ' . $this->tableName . ' WHERE lower(email)=lower(?)';
+                $stmt = $this->pdo->prepare($query);
+                $stmt->execute([ $value ]);
+            
+                while ($row = $stmt->fetch(\PDO::FETCH_ASSOC)) {
+                       // Checking if the principal is in the prefix
+                       list($rowPrefix) = URLUtil::splitPath($row['uri']);
```
staabm: indent seems to be off

evert:
Closing this in favor of #749

PR #703, opened 2015-08-11
fix client prop patch should add {DAV:}prop parent node

jeremydenoun:
After some test, it's seems OOB Sabre WebDav client Propatch method not correctly encapsulate element in XML request (miss '{DAV:}prop') I propose patch with this element


evert:
Thanks for the patch!

PR #707, opened 2015-08-26
SupportedCalendarComponentSet deserializer fails for empty/no values

DominikTo:
* Added SupportedCalendarComponentSetTest

### DIFF
--- a/tests/Sabre/CalDAV/Xml/Property/SupportedCalendarComponentSetTest.php
+++ b/tests/Sabre/CalDAV/Xml/Property/SupportedCalendarComponentSetTest.php
@@ -1,0 +1,104 @@
+<?php
+
+namespace Sabre\CalDAV\Xml\Property;
+
+use Sabre\CalDAV;
+use Sabre\DAV;
+
+class SupportedCalendarComponentSetTest extends DAV\Xml\XmlTest {
+
+    function setUp() {
+
+        $this->namespaceMap[CalDAV\Plugin::NS_CALDAV] = 'cal';
+        $this->namespaceMap[CalDAV\Plugin::NS_CALENDARSERVER] = 'cs';
+
+    }
+
+    function testSimple() {
+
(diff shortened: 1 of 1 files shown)


DominikTo:
`parseInnerTree()` returns `null` here. We could always cast the result from [`$reader->parseInnerTree()`](https://github.com/fruux/sabre-dav/blob/cd4e193e5fbaee33eeaf0e3e6e5945c43e0fa289/lib/CalDAV/Xml/Property/SupportedCalendarComponentSet.php#L110) to an array or only run the loop if `$elems` is an array.

There are two more occurrences that are probably affected by this:

* [`Sabre\CalDAV\Xml\Property\Invite`](https://github.com/fruux/sabre-dav/blob/cd4e193e5fbaee33eeaf0e3e6e5945c43e0fa289/lib/CalDAV/Xml/Property/Invite.php#L193)
* [`Sabre\CalDAV\Xml\Property\ScheduleCalendarTransp`](https://github.com/fruux/sabre-dav/blob/cd4e193e5fbaee33eeaf0e3e6e5945c43e0fa289/lib/CalDAV/Xml/Property/ScheduleCalendarTransp.php#L123)

evert:
One question I would have though... why is it empty? A bug elsewhere?

evert:
The specification actually says that it needs to contain at least 1 sub-component. So I'm changing the test to expect an exception and then fix to code to emit that =)

DominikTo:
:+1:

Does that also apply to `Invite` and `ScheduleCalendarTransp` (see above) that there should never be a case where we run the foreach with a `null` value? At least for `Invite`, we've seen this in fruux/webapp#370.

evert:
Yea true... Just added a new function in sabre/xml to generalize this case. Once sabre/dav updates to sabre/xml 1.2, I can just mass-replace `parseInnerTree` with `parseGetElements`.

https://github.com/fruux/sabre-xml/issues/39

PR #710, opened 2015-08-30
Add HTTP Bearer Auth support

fkooman:


composer.json
```diff
@@ -13,12 +13,18 @@
             "role" : "Developer"
         }
     ],
+    "repositories": [
```
evert: This needs to go =)

composer.json
```diff
@@ -13,12 +13,18 @@
...
+            "url": "https://github.com/fkooman/sabre-http"
+        }
+    ],
     "require": {
         "php": ">=5.5.0",
         "sabre/vobject": ">=4.0.0-alpha1",
         "sabre/event" : "~2.0",
         "sabre/xml"  : "~1.0",
-        "sabre/http" : "~4.0",
+        "sabre/http" : "dev-bearer",
```
DominikTo: `Sabre\HTTP\Auth\Bearer` has been added to sabre/http 4.1.0.

lib/DAV/Auth/Backend/AbstractBearer.php
```diff
@@ -0,0 +1,145 @@
...
+     * @var string
+     */
+    protected $principalPrefix = 'principals/';
+
+    /**
+     * Validates a Bearer token
+     *
+     * This method should return true or false depending on if login
+     * succeeded.
+     *
```
evert: Docblock is wrong. This method must return false or a principal path.

lib/DAV/Auth/Backend/AbstractBearer.php
```diff
@@ -0,0 +1,145 @@
...
+     * @var string
+     */
+    protected $realm = 'sabre/dav';
+
+    /**
+     * This is the prefix that will be used to generate principal urls.
+     *
+     * @var string
+     */
+    protected $principalPrefix = 'principals/';
```
evert: Take out principal prefix. This only really makes sense for username-based authentication. Instead `validateBearerToken` should return the full principal path.

evert:
I have a last question... do we need the realm? It might be better to take it out completely. It's required for digest and basic, but not here. Does it serve a strong purpose, or would it be 'just as good' without for most practical purposes?

fkooman:
About the realm, see http://tools.ietf.org/html/rfc6750

```
A "realm" attribute MAY be included to indicate the scope of
protection in the manner described in HTTP/1.1 [RFC2617].  The
"realm" attribute MUST NOT appear more than once.
```

In RFC 2617:

```
The realm directive (case-insensitive) is required for all
authentication schemes that issue a challenge.
```

So it is also not needed for Basic there?

I left it in because it is in all the examples in the RFC 6750 document. Is there a good reason to remove it?

lib/DAV/Auth/Backend/AbstractBearer.php
```diff
@@ -0,0 +1,138 @@
...
+
+        $auth = new HTTP\Auth\Bearer(
+            $this->realm,
+            $request,
+            $response
+        );
+
+        $bearerToken = $auth->getToken($request);
+        if (!$bearerToken) {
+            return [false, "No 'Authorization: Bearer' header found. Either the client didn't send one, or the server is mis-configured"];
```
LukasReschke: `mis-configured` looks strange: http://www.oxforddictionaries.com/definition/english/misconfigure

PR #712, opened 2015-09-02
Prevent MOVE from overwriting destination if source can not be deleted

### LINKED ISSUE #705: Make it possible to disable deletion on MOVE + overwrite
Consider the current scenario:

1. a file exists in "/a/test.txt" but the user only has read permissions
2. another file exists as "/test.txt"
3. the user performs a MOVE "/a/test.txt" "/test.txt" with overwrite

In the normal case of MOVE + overwrite, Sabre will automatically delete the target file "/test.txt".
But here the user isn't allowed to move the source file.

It should be possible to tell Sabre to not automatically delete the target file and leave the decision up to the implementation of $tree->move() to decide whether to delete the target or not.

Also the `beforeMove` hook is fired too late: https://github.com/fruux/sabre-dav/blob/master/lib/DAV/CorePlugin.php#L636

The only way I can think of is overriding the Server class / Core plugin somehow and overriding either `httpMove` or `getCopyAndMoveInfo` and return false for "destinationExists".

What do you think ?

evert:
This PR is supposed to fix #705.


tests/Sabre/DAV/Mock/Collection.php
```diff
@@ -95,7 +105,7 @@ public function createFile($name, $data = null) {
      */
     public function createDirectory($name) {
 
-        $this->children[$name] = array();
+        $this->children = new Collection($name);
```
staabm: in line 40 this property is handled as an array. here as a scalar property...=

lib/DAV/CorePlugin.php
```diff
@@ -633,13 +633,18 @@ public function httpMove(RequestInterface $request, ResponseInterface $response)
         if ($moveInfo['destinationExists']) {
 
             if (!$this->server->emit('beforeUnbind',[$moveInfo['destination']])) return false;
+
+        }
+        if (!$this->server->emit('beforeUnbind',[$path])) return false;
```
staabm: wouldn't it emit the `beforeUnbind` event twice now? intended?

evert: There's on for the source, and one for the destination.

evert:
Closing this PR because the incorrect branch was targetted.

PR #714, opened 2015-09-04
Prevent delete on forbidden move

### LINKED ISSUE #705: Make it possible to disable deletion on MOVE + overwrite

evert:
This PR is supposed to fix #705.


Hywan:
A little `git rebase` to fixup/squash may be required before merging :-].

evert:
Just made the history a bit cleaner.

PR #729, opened 2015-10-25
Do not prepend / if the basepath is empty

rullzer:
Described in https://github.com/owncloud/core/pull/20003#issuecomment-150934856

Basically in the root the basepath is empty. And this childnodes are cached like '/<CHILD>' while they should be cached as '<CHILD>'.

It seems this is also still the case in the current master.

### DIFF
--- a/lib/DAV/Tree.php
+++ b/lib/DAV/Tree.php
@@ -193,7 +193,8 @@ class Tree {
 
         $node = $this->getNodeForPath($path);
         $children = $node->getChildren();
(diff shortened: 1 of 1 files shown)


evert:
Hey @rullzer , thanks for finding this. I wrote an alternative patch for this in 6223d06b

Your own patch has an issue with strings such as `'0'`.

Good find though!

rullzer:
Ah yes always fun that '0' stuff in php. :wink:

PR #740, opened 2016-01-03
Fix errors when running migration with MySQL

pageer:
This fixes two errors that cause the migration to throw an exception and fail.  The first is a case fall-through that causes the SQLite case to be executed as well and the second is a syntax error in a query.


evert:
Going to backport this

PR #749, opened 2016-01-05
Better findByUri 

evert:
This is an updated version of #686

### DIFF
--- a/lib/DAVACL/PrincipalBackend/PDO.php
+++ b/lib/DAVACL/PrincipalBackend/PDO.php
@@ -240,26 +240,25 @@ class PDO extends AbstractBackend implements CreatePrincipalSupport {
      * @return array
      */
     function searchPrincipals($prefixPath, array $searchProperties, $test = 'allof') {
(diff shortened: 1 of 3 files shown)


lib/DAVACL/PrincipalBackend/PDO.php
```diff
@@ -281,6 +280,52 @@ function searchPrincipals($prefixPath, array $searchProperties, $test = 'allof')
...
+    function findByUri($uri, $principalPrefix) {
+        $value = null;
+        $scheme = null;
+        list($scheme, $value) = explode(":", $uri, 2);
+        if ($value == null) return null;
+
+        $uri = null;
+        switch ($scheme){
+            case "mailto":
+                $query = 'SELECT uri FROM ' . $this->tableName . ' WHERE lower(email)=lower(?)';
```
staabm: You should define the column with a `_ci` collation and do direct fomparions without `lower` so the dh can utilise its index

evert: sounds like a good idea =)

PR #750, opened 2016-01-05
FIX: $propPatch->commit(); is missing in updatePrincipal() method for…

dapepe:
… Sabre\DAVACL\PrincipalBackend\PDO

### DIFF
--- a/lib/DAVACL/PrincipalBackend/PDO.php
+++ b/lib/DAVACL/PrincipalBackend/PDO.php
@@ -208,6 +208,8 @@ class PDO extends AbstractBackend implements CreatePrincipalSupport {
 
         });
 
+        $propPatch->commit();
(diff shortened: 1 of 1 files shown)


evert:
Hi!

This is expected, because there normally needs to something *outside* of that method that initiates the commit. The reason for this is that a `PROPPATCH` request potentially has a set of properties that need to be handled by different subsystems (such as the PrincipalBackend but also for instance the PropertyStorage backend).

Only when all properties within a single `PROPPATCH` are "handled", the system will call commit.

The underlying reason is that we might call multiple backends to handle the request, and if the first succeeds, and second fails we end up with an incorrect state.

dapepe:
Understood! Thanks for the fast feedback! :-)

PR #772, opened 2016-02-03
Various style fixes for CalDav

jakobsack:
- remove unneeded variables
- stop tests right after the point where they should fail
- upcase names of VObject magic attributes
- remove unnecessary parameters
- add missing parameters


evert:
Thank you!

PR #774, opened 2016-02-07
Relative paths won't work on Windows

TonyBogdanov:
Seems that disk_total_space and disk_free_space cannot work with relative paths on Windows (WAMP + PHP 5.6)

### DIFF
--- a/lib/DAV/FS/Directory.php
+++ b/lib/DAV/FS/Directory.php
@@ -140,7 +140,7 @@ class Directory extends Node implements DAV\ICollection, DAV\IQuota {
      * @return array
      */
     function getQuotaInfo() {
-        $absolute = getcwd() . '/' . $this->path;
+        $absolute = realpath($this->path);
         return [
             disk_total_space($absolute) - disk_free_space($absolute),
             disk_free_space($absolute)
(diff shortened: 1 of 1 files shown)


lib/DAV/FS/Directory.php
```diff
@@ -140,10 +140,10 @@ function delete() {
      * @return array
      */
     function getQuotaInfo() {
-
+        $absolute = getcwd() . '/' . $this->path;
```
staabm: If relative paths are the issue, you should not assume the cwd, but use `realpath`

staabm:
Do you get any errors/warnings without this patch?

TonyBogdanov:
The issue seems to appear only on Windows when PHP is used with Apache. When run in the console (with the same `php.ini` apache's using) it's fine.

Otherwise I get `Warning: disk_total_space(): The system cannot find the path specified.`.

I'll try and report this to php, for the time being, `realpath()` should do the trick, even travis agrees :)

evert:
I would definitely _not_ run sabre/dav on windows. Pretty bad idea unless you _only_ deal with ascii filenames.

That being said, I'm happy with this patch =)

PR #782, opened 2016-02-17
update the year in Browser\plugin.php

kralo:
not only for aesthetics, but this way you don't get confused when you just installed the latest version and wonder about the year


kralo:
and btw: this HTML-Browsing of the DAV internals is wonderful  :sunglasses:

evert:
Can't believe I forgot!

Issue #787, opened 2016-02-24
[regression] Cannot use non-seekable stream wrappers with range requests


PVince81:
In Sabre 2.1, the CorePlugin was checking the result of `fseek` when using range requests.
If fseek failed, it would read enough bytes to reach the correct position.
https://github.com/fruux/sabre-dav/blob/2.1/lib/DAV/CorePlugin.php#L177

In Sabre 3, the result of fseek is not checked for errors any more: https://github.com/fruux/sabre-dav/blob/3.0/lib/DAV/CorePlugin.php#L176
A `stream_get_meta_data` checks the "seekable" property of the stream.

The problem is that stream wrappers always advertise themselves as seekable, even when the wrapped stream isn't. This is what we observed and was also observed by other people here http://php.net/manual/en/function.stream-get-meta-data.php#115608

It seems there is no way for a stream wrapper to control what data is returned by `stream_get_meta_data`, so using this cannot be used to safely check for seekability.

Please revert back or add a check for the `fseek` call.

@evert

PVince81:
Ref https://github.com/owncloud/core/issues/22476 in which case the wrapped stream is a Guzzle response (curl).

staabm:
could we detect whether the stream is provided using a custom-wrapper?

PVince81:
Even if you could, how would you change the code to make it work ?
There are situations where the custom wrapper is wrapping a seekable stream and fseek will work.
In other situations where the underlying stream is not seekable, fseek would not work.

evert:
I think just calling fseek and see if it did anything does indeed sound the sanest. Sucks that streams is in such a poor state in PHP.

PR #789, opened 2016-02-28
Fall back to manually seeking if fseek fails.

evert:
This is a potential fix for #787. This targets the 3.0 branch and should
then get merged into the 3.1 and master branches.

@PVince81: How does this look?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ ChangeLog
 
 * #784: Sync logs for address books were not correctly cleaned up after
   deleting them.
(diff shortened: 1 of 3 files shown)


evert:
Travis doesn't like it.

evert:
Corrected the bug. Wanna take a look @PVince81 ?

PVince81:
Thanks a lot! I'll have a look.

PVince81:
Tested the patch against Sabre 3.0, it worked fine. The block of code that eats bytes was correctly triggered. :+1:

PR #795, opened 2016-03-14
drop hhvm from travis since underlying sabre-xml doesnt support it ei…

staabm:
…ther

as long as sabre-xml does not properly support hhvm we shouldnt test it either.
IMO sabre-dav cannot work on hhvm as long as all the dependencies dont support it.


evert:
It's in allowed failures. I think it's helpful to keep an eye on it

PR #796, opened 2016-03-14
download composer dependencies in parallel

staabm:
testing... do not merge yet


evert:
Not a big fan of a new dependency for this to be honest =)

staabm:
Its just a perf related change (dev only). Since the build doesnt take hours I am fine with closing it unmerged.

PR #802, opened 2016-03-24
Fixed hhvm error caused by xdebug disable procedure

staabm:


evert:
The `; true` trick actually worked!

Issue #808, opened 2016-04-01
Add strict validation / repair of vCards and iCalendar when receiving them via PUT


evert:
In the past, when we receive a new vCard or iCalendar object, we just do a small amount of validation before storing the objects.

Basically the validation is simply: is vobject able to parse it.

When parsing vobject _only_ does a structural check of the object, and doesn't automatically validate the individual properties for correctness. This would be similar to checking "is the xml document valid xml" without ensuring that the individual elements also match what you expect.

This has caused a number of issues, especially as we're getting more strict over time. In particular, an object may have been valid in a previous version, and no longer in a newer version, which trips up several `REPORT` requests.

While it's hard to fix past versions and the data they already accepted, we can do better for future objects.

I'm now adding a feature that:

1. Calls Vobject::validate() for any incoming object.
2. Automatically repairs problems, if possible.
3. If not, throw a `400 Bad Request`.
4. If it _was_ possible to do repairs, we _don't_ return an Etag.
5. But if the object was already valid, we _do_ return an ETag.
6. If the client supplies a `Prefer: handling=strict` header we _never_ attempt to do a repair.

It is possible that misbehaving clients will stop functioning after that, but that's a good thing. It allows us to report bugs for those clients, or create more 'fixers' in vobject, if possible.

Fixes #421

PR #809, opened 2016-04-01
Improve validation

evert:
this is a wip. Fixing #808 

Todo:

**vCards**

* [x] Automatically repair, if possible
* [x] Throw error when not possible.
* [x] Add support for `Prefer: handling=strict` HTTP header to ask server for 'strict mode'

**iCalendar**

* [x] Automatically repair, if possible
* [x] Throw error when not possible.
* [x] Add support for `Prefer: handling=strict` HTTP header to ask server for 'strict mode'

### DIFF
--- a/lib/CardDAV/Plugin.php
+++ b/lib/CardDAV/Plugin.php
@@ -361,9 +361,14 @@ class Plugin extends DAV\ServerPlugin {
             throw new DAV\Exception\UnsupportedMediaType('This collection can only support vcard objects.');
         }
 
-        $messages = $vobj->validate(
-            VObject\Node::PROFILE_CARDDAV | VObject\Node::REPAIR
-        );
+        $options = VObject\Node::PROFILE_CARDDAV;
+        $prefer = $this->server->getHTTPPrefer();
+
+        if ($prefer['handling'] !== 'strict') {
+            $options |= VObject\Node::REPAIR;
+        }
+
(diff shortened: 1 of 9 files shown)


lib/CardDAV/Plugin.php
```diff
@@ -366,11 +361,56 @@ protected function validateVCard(&$data, &$modified) {
...
+                    break;
+                case 3 :
+                    // Level 3 means a critical error
+                    throw new DAV\Exception\UnsupportedMediaType('Validation error in vCard: ' . $message['message']);
+
+            }
+
+        }
+        if ($warningMessage) {
+            $this->server->httpResponse->setHeader(
```
staabm: Debug code?

evert: It was actually intentional. `Ew` seemed appropriate, because we accepted the vCard, but we think its a little gross. Maybe `Ew-Gross` is better.

Afaik there's no good standard HTTP header for this.

staabm: if this is a app/lib specific header, should it be `X-...`?

evert: Not a bad idea. Officially the IETF now recommends against x-headers, but the alternative is to formally register it. Might as well be consistent

evert:
Ready for review

PR #812, opened 2016-04-03
Set a filename with .ics extension when exporting a calendar

Xenopathic:
Format is `<calendar>-YYYY-MM-DD.ics`, or `<calendar>-YYYY-MM-DD.json` if using the jCal export format. I included a date, since typically a calendar export is done as a backup or 'snapshot' of the current calendar, in which it is almost always useful to have the file tagged with the date of creation.

Downstream feature request: owncloud/core#23751

### DIFF
--- a/lib/CalDAV/ICSExportPlugin.php
+++ b/lib/CalDAV/ICSExportPlugin.php
@@ -170,13 +170,13 @@ class ICSExportPlugin extends DAV\ServerPlugin {
     protected function generateResponse($path, $start, $end, $expand, $componentType, $format, $properties, ResponseInterface $response) {
 
         $calDataProp = '{' . Plugin::NS_CALDAV . '}calendar-data';
+        $calendarNode = $this->server->tree->getNodeForPath($path);
 
         $blobs = [];
         if ($start || $end || $componentType) {
 
             // If there was a start or end filter, we need to enlist
             // calendarQuery for speed.
-            $calendarNode = $this->server->tree->getNodeForPath($path);
             $queryResult = $calendarNode->calendarQuery([
                 'name'         => 'VCALENDAR',
                 'comp-filters' => [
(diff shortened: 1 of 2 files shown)


evert:
Thanks for a very complete PR!

lib/CalDAV/ICSExportPlugin.php
```diff
@@ -246,17 +246,30 @@ function($item) use ($path) {
...
                 break;
             case 'application/calendar+json' :
                 $mergedCalendar = json_encode($mergedCalendar->jsonSerialize());
+                $filenameExtension = '.json';
                 break;
         }
 
+        $filename = null;
+        if (isset($properties['{DAV:}displayname']) && $properties['{DAV:}displayname'] !== '') {
+            $filename = $properties['{DAV:}displayname'];
```
evert: My main concern is: could this be a security issue? This line would allow a user to inject any string into a http header / file name.

You could for instance add a `"` to the filename, to escape from the double quotes in the header. I'd wonder if there's a way we can make this safer.

staabm: I guess easiest would be to preg-replace with a white-list

Xenopathic:
@evert @staabm I've put a filter in for the most common characters (and, ofc, a unit test)

lib/CalDAV/ICSExportPlugin.php
```diff
@@ -246,17 +246,34 @@ function($item) use ($path) {
...
+                $filenameExtension = '.ics';
                 break;
             case 'application/calendar+json' :
                 $mergedCalendar = json_encode($mergedCalendar->jsonSerialize());
+                $filenameExtension = '.json';
                 break;
         }
 
+        $filename = null;
+        if (isset($properties['{DAV:}displayname']) && $properties['{DAV:}displayname'] !== '') {
```
DeepDiver1975: Why use the displayname? I'd go for the name of the calendar.

If I do a 
```
wget http://example.com/dav/calendars/tom/my-own-private-calendar.ics
```

I'd expect there a file to be downloaded with the name my-own-private-calendar.ics

Xenopathic: Short of splitting the path URL, there's no handy property for that. In many generic cases though, where the URL doesn't correspond to any sensible name (like some sort of UUID for example), using this display name is better IMO. Perhaps @evert has an idea?

DeepDiver1975: The code is operating on basis of a calendar node and like any other node we can call getName() on it which will return the uri

evert: Frankly I believe that both using the name or the displayname has merit. Can you guys decide? =)

Xenopathic: I originally wanted to use the name of the calendar, but switched to the display name since I couldn't figure out an easy way to get the name. @DeepDiver1975 has pointed me in the right direction though, so I'll change it :smile:

PR #814, opened 2016-04-10
Set filename with .vcf extension for address book exports

Xenopathic:
Very similar to #812, but for CardDAV's VCFExportPlugin instead.

cc @skjnldsv @DeepDiver1975


skjnldsv:
:+1:

PR #816, opened 2016-04-11
Postgres update

plockaby:
While running against postgres, which I know is unsupported, it started to come up where the "transparent" column was getting set to an empty value that either PDO or Postgres wasn't liking. This change made everything work again.


evert:
Thank you! looks good

PR #820, opened 2016-04-18
Added support for ACL-based unauthenticated access.

### LINKED ISSUE #657: Disable auth if `{DAV:}unauthenticated`
When ACL contains `{DAV:}unauthenticated` or `{DAV:}all` we should allow the ability to perform these actions without requiring authentication.

The way this needs to work is as follows:

1. The authentication plugin now automatically calls `challenge` on every authentication backend. The plugin should get some kind of flag that allows someone to disable this. Lets call it `noAutoChallenge` for the moment.
2. The acl plugin should automatically look for the authentication plugin, and set the `noAutoChallenge` to true.
3. When the ACL plugin is required to check for privileges (`checkPrivileges` method), and the check _fails_, check if *any* user is currently logged in. If no user is logged in, call the `challenge` method manually on the authentication plugin.

In short: we postpone the challenge *until* the first time an ACL check fails.

Risks
-------

The risk is that *before* this change, *any* action would require authentication, and after this change this is no longer true. There's a possibility that there's things within sabre/dav that don't correctly check the access control system before doing something.

If this is true, we now open up these requests, allowing people to perform these actions without requiring proper authentication. This could be bad.

Even if sabre/dav code doesn't do this, other people that use sabre/dav might.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/15947002-disable-auth-if-dav-unauthenticated?utm_campaign=plugin&utm_content=tracker%2F294006&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F294006&utm_medium=issues&utm_source=github).
</bountysource-plugin>

evert:
This PR adds support for `{DAV:}unauthorized` in ACE rules.

The implication is that privileges can be assigned to users that are not currently logged in. This PR works by temporarily turning off the authentication plugin, and only force a user to authenticate if they are attempting to perform an operation they don't have access to.

Fixes #657.

### DIFF
--- a/lib/DAV/Auth/Plugin.php
+++ b/lib/DAV/Auth/Plugin.php
@@ -25,6 +25,20 @@ use Sabre\DAV\ServerPlugin;
 class Plugin extends ServerPlugin {
 
     /**
+     * By default this plugin will require that the user is authenticated,
+     * and refuse any access if the user is not authenticated.
+     *
+     * If this setting is set to false, we let the user through, whether they
+     * are authenticated or not.
+     *
+     * This is useful if you want to allow both authenticated and
+     * unauthenticated access to your server.
+     *
+     * @param bool
+     */
+    public $autoRequireLogin = true;
+
+    /**
      * authentication backends
      */
     protected $backends;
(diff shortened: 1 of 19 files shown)


Hywan:
\o/

Hywan:
This will be very useful for the “public directory” of sabre/katana.

lib/DAV/Auth/Plugin.php
```diff
@@ -132,6 +146,50 @@ function beforeMethod(RequestInterface $request, ResponseInterface $response) {
...
+
+        $authResult = $this->check($request, $response);
+
+        if ($authResult[0]) {
+            // Auth was successful
+            $this->currentPrincipal = $authResult[1];
+            $this->loginFailedReasons = null;
+            return;
+        }
+
```
Hywan: Another newline is missing :-p?

evert: Lol. Some kind of diss?

Issue #829, opened 2016-05-02
addressbook-multiget on not existing resources causes the whole request to fail with 404 - shouldn't that be in the multi status response?


DeepDiver1975:
see https://github.com/owncloud/core/issues/24193

I can simply reporduce this be sending an addressbook multiget request to the server asking for vcard which do not exist (e.g. having an invalid principal or addressbook name)
![bildschirmfoto von 2016-04-25 11-33-32](https://cloud.githubusercontent.com/assets/1005065/14779793/055e6650-0ada-11e6-9109-ba8143ee1cbe.png)

In case the vcard name itself is not available in the addressbook no result is returned:
![bildschirmfoto von 2016-04-25 11-32-54](https://cloud.githubusercontent.com/assets/1005065/14779665/7d3162c8-0ad9-11e6-81fd-af1b36747a07.png)

evert:
You're right, and this likely also needs to be corrected in caldav.

DeepDiver1975:
If you agree I can work on a PR for this ... for master and also 3.0.x

DeepDiver1975:
@evert  Sabre\DAV\Xml\Response\MultiStatus cannot handle a structure like this yet - right?

see https://tools.ietf.org/html/rfc6352#section-8.7.1

```xml
 <?xml version="1.0" encoding="utf-8" ?>
   <D:multistatus xmlns:D="DAV:"
                  xmlns:C="urn:ietf:params:xml:ns:carddav">
     <D:response>
       <D:href>/home/bernard/addressbook/vcf1.vcf</D:href>
       <D:status>HTTP/1.1 404 Resource not found</D:status>
     </D:response>
   </D:multistatus>

evert:
Yea in cases where I needed it, I just manually constructed `Response` objects, such as in the respnonse for sync-collection.

PR #830, opened 2016-05-03
Handle 404 in CardDAV multi get

DeepDiver1975:
refs #829

### DIFF
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -1017,20 +1017,25 @@ class Server implements LoggerAwareInterface, EmitterInterface {
         $result = [
         ];
 
-        $nodes = $this->tree->getMultipleNodes($paths);
+        foreach ($paths as $path) {
 
-        foreach ($nodes as $path => $node) {
+            try {
+                $node = $this->tree->getNodeForPath($path);
+                $propFind = new PropFind($path, $propertyNames);
+                $r = $this->getPropertiesByNode($propFind, $node);
+                if ($r) {
+                    $result[$path] = $propFind->getResultForMultiStatus();
+                    $result[$path]['href'] = $path;
 
-            $propFind = new PropFind($path, $propertyNames);
-            $r = $this->getPropertiesByNode($propFind, $node);
-            if ($r) {
-                $result[$path] = $propFind->getResultForMultiStatus();
-                $result[$path]['href'] = $path;
-
-                $resourceType = $this->getResourceTypeForNode($node);
-                if (in_array('{DAV:}collection', $resourceType) || in_array('{DAV:}principal', $resourceType)) {
-                    $result[$path]['href'] .= '/';
+                    $resourceType = $this->getResourceTypeForNode($node);
+                    if (in_array('{DAV:}collection', $resourceType) || in_array('{DAV:}principal', $resourceType)) {
+                        $result[$path]['href'] .= '/';
+                    }
                 }
+            } catch (Exception\NotFound $ex) {
+                $result[$path] = [];
+                $result[$path]['href'] = $path;
+                $result[$path]['status'] = 404;
             }
 
         }
@@ -1670,9 +1675,11 @@ class Server implements LoggerAwareInterface, EmitterInterface {
             if ($strip404s) {
                 unset($entry[404]);
             }
+            $status = isset($entry['status']) ? $entry['status'] : null;
             $response = new Xml\Element\Response(
                 ltrim($href, '/'),
-                $entry
+                $entry,
+                $status
             );
             $w->write([
                 'name'  => '{DAV:}response',
--- a/lib/DAV/Xml/Element/Response.php
+++ b/lib/DAV/Xml/Element/Response.php
@@ -27,7 +27,7 @@ class Response implements Element {
     protected $href;
 
     /**
-     * Propertylist, ordered by HTTP status code
+     * Property list, ordered by HTTP status code
      *
      * @var array
      */
@@ -119,7 +119,7 @@ class Response implements Element {
      */
     function xmlSerialize(Writer $writer) {
 
-        if ($status = $this->getHTTPStatus()) {
+        if ($status = $this->getHttpStatus()) {
             $writer->writeElement('{DAV:}status', 'HTTP/1.1 ' . $status . ' ' . \Sabre\HTTP\Response::$statusCodes[$status]);
         }
         $writer->writeElement('{DAV:}href', $writer->contextUri . \Sabre\HTTP\encodePath($this->getHref()));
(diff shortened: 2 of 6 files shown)


DeepDiver1975:
@evert ready to go from my pov - let me know what you think - THX

DeepDiver1975:
I'll take care about CalDAV multiget once we agree on this being the right way to do it

lib/DAV/Server.php
```diff
@@ -993,20 +994,26 @@ function getPropertiesForMultiplePaths(array $paths, array $propertyNames = [])
...
+                    $result[$path]['href'] = $path;
+
+                    $resourceType = $this->getResourceTypeForNode($node);
+                    if (in_array('{DAV:}collection', $resourceType) || in_array('{DAV:}principal', $resourceType)) {
+                        $result[$path]['href'] .= '/';
+                    }
                 }
+            } catch (NotFound $ex) {
+                $result[$path] = [
+                    404 => [],
```
evert: I'm not sure about this response status. This tells me that the path is found, but it had 0 properties.

So two ideas:

1. Set `$result[$path] = null` for items that are not found.
2. Add a new `status` property to every item in `$result`, indicating the "overall" status of the thing on that path.

Either works for me, just `null` is a good indicator that the thing didn't exist and makes it very obvious, whereas adding a `status` would allow us to handle errors other than 404's in the future.

lib/DAV/Xml/Element/Response.php
```diff
@@ -119,25 +119,25 @@ function getResponseProperties() {
...
+            if ($properties && (ctype_digit($status) || is_int($status))) {
+                $empty = false;
+                $writer->startElement('{DAV:}propstat');
+                $writer->writeElement('{DAV:}prop', $properties);
+                $writer->writeElement('{DAV:}status', 'HTTP/1.1 ' . $status . ' ' . \Sabre\HTTP\Response::$statusCodes[$status]);
+                $writer->endElement(); // {DAV:}propstat
+            }
+            if ($empty && !$properties && (ctype_digit($status) || is_int($status))) {
+                $empty = false;
+                $writer->writeElement('{DAV:}status', 'HTTP/1.1 ' . $status . ' ' . \Sabre\HTTP\Response::$statusCodes[$status]);
```
evert: I also think none of these changes should really be needed. We're already witing that `{DAV:}status` element a few lines up

DeepDiver1975: The status above is for the full response but we need the status down here for the individual elements.

DeepDiver1975:
@evert please have a look - THX

lib/DAV/Xml/Element/Response.php
```diff
@@ -119,13 +119,15 @@ function getResponseProperties() {
      */
     function xmlSerialize(Writer $writer) {
 
-        if ($status = $this->getHTTPStatus()) {
+        $empty = true;
+
+        if ($status = $this->getHttpStatus()) {
             $writer->writeElement('{DAV:}status', 'HTTP/1.1 ' . $status . ' ' . \Sabre\HTTP\Response::$statusCodes[$status]);
+            // in case of a 404 we simply don't expect any properties at all
+            $empty = ($status !== 404);
```
evert: I think I don't like this line. Could it get removed?

I don't think realistically that there might be a case where there's a resource-level 404 status, where we're returning properties anyway, but I don't really see a reason to special-case this.

evert:
Great patch otherwise :+1:

lib/DAV/Xml/Element/Response.php
DeepDiver1975: If we don't set empty to true we will get a propset with 418/teapot inside. Kind of strange to me.

lib/DAV/Xml/Element/Response.php
```diff
@@ -137,7 +139,6 @@ function xmlSerialize(Writer $writer) {
             $writer->writeElement('{DAV:}prop', $properties);
             $writer->writeElement('{DAV:}status', 'HTTP/1.1 ' . $status . ' ' . \Sabre\HTTP\Response::$statusCodes[$status]);
             $writer->endElement(); // {DAV:}propstat
-
         }
         if ($empty) {
```
evert: @DeepDiver1975 ah then this block really should be `$empty && !$status` all along. Preferably this function should be rewritten a bit so there's no duplication in writing the status element.

evert: poor coding on my side. Might have been drunk. Would explain the teapot

DeepDiver1975: Have some more :beer:

DeepDiver1975: There is one Test here which expects an explicit status of 200 and an teaport propset.

I'll post it in a bit.

DeepDiver1975:
```sh
There was 1 failure:

1) Sabre\DAVACL\PrincipalMatchTest::testPrincipalMatch
Failed asserting that two DOM documents are equal.
--- Expected
+++ Actual
@@ @@
 <?xml version="1.0"?>
 <d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
   <d:status>HTTP/1.1 200 OK</d:status>
   <d:href>/principals/user1</d:href>
-  <d:propstat>
-    <d:prop/>
-    <d:status>HTTP/1.1 418 I'm a teapot</d:status>
-  </d:propstat>
 </d:multistatus>

**Comment by DeepDiver1975** [comment]
````
There was 1 failure:

1) Sabre\DAVACL\PrincipalMatchTest::testPrincipalMatch

Failed asserting that two DOM documents are equal.

--- Expected

+++ Actual

@@ @@

 <?xml version="1.0"?>

 <d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">

   <d:status>HTTP/1.1 200 OK</d:status>

   <d:href>/principals/user1</d:href>

-  <d:propstat>

-    <d:prop/>

-    <d:status>HTTP/1.1 418 I'm a teapot</d:status>

-  </d:propstat>

 </d:multistatus>

/home/travis/build/fruux/sabre-dav/tests/Sabre/DAVACL/PrincipalMatchTest.php:41

**Comment by DeepDiver1975** [comment]
@evert Can I drag your attention onto this one?

**Comment by DeepDiver1975** [comment]
@evert all green now - please have a second look. THX

File: lib/DAV/Server.php [hunk 5]
```diff
@@ -994,20 +995,25 @@ function getPropertiesForMultiplePaths(array $paths, array $propertyNames = [])
         $result = [
         ];
 
-        $nodes = $this->tree->getMultipleNodes($paths);
+        foreach ($paths as $path) {
```
evert: I'm super sorry for not spotting this issue earlier. I'm not as focused as I used to... but I see a new problem :(  

The new problem with this patch, that I didn't see before, is that you're now no longer using `getMultipleNodes()`. While this solves the problem, the new problem is that that function was actually implemented to solve a fairly large performance problem.

There's several requests (such as sync-collection and multiget reports) that need to do a *lot* of individual lookups, and by adding the "Multiple" interfaces, new optimizations were possible.

So what can be done? I think the best thing is that getMultipleNodes actually gets changed to catch errors. The simplest solution there would be to omit (or return null) for the paths that throw a NotFound.

Then any non-NotFound exceptions will still 'fatal' though, so if we handle those as well in this patch we need a different mechanism. Perhaps the result of `getMultipleNodes` could return either a `Node` or an instance of `Exception` in case there was a problem? That would at least give us all the information (although it's a bit unconventional).

Sorry again :/

PR #845, opened 2016-05-30
IMipPlugin :: add sender email / check recipient domain / simple text

torbbe:
…xt abstract

Changes, which maybe helpful
- Use the sender email address of the invitation instead the no-reply
- Check domain of the recipient, if is the same like the given domain, no email is send (reduce email inside one domain) > invitation is provided by schedule plugin!?
- not just send the calendar object, add a simple text abstract of the invitation

### DIFF
--- a/lib/CalDAV/Schedule/IMipPlugin.php
+++ b/lib/CalDAV/Schedule/IMipPlugin.php
@@ -200,12 +200,12 @@ class IMipPlugin extends DAV\ServerPlugin {
     function getMessage($iTipMessage) {
 
         // Text message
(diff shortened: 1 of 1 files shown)


DeepDiver1975:
From my understanding to is already a too specific implementation. We should provide a way to let developers to compose mails as they like.

As of today inheriting the IMipPlugin leads to reimplementing the scheduling logic - see https://github.com/owncloud/core/blob/3a9b9dda0eea3d3ea3dfdeb9721ad892b1f7e27a/apps/dav/lib/CalDAV/Schedule/IMipPlugin.php

Let's see if I can come up with something .....

PR #849, opened 2016-06-10
Handle 404 with no PROPSTAT

cedtaz:
When i delete a contact on addressbook server, the action doesn't appear on Thunderbird with plugin Sogo.
Here are the comments in the source code of Sogo : 
/* 404 responses are now supposed to occur only when no propfind is present. Yet, the "status" seems not mandatory so we play it safe here. */
So, the 404 response with PROPSTAT was not considered.

### DIFF
--- a/lib/DAV/Xml/Element/Response.php
+++ b/lib/DAV/Xml/Element/Response.php
@@ -148,10 +148,12 @@ class Response implements Element {
              * In those cases we MUST specify at least one DAV:propstat anyway, with
              * no properties.
              */
-            $writer->writeElement('{DAV:}propstat', [
-                '{DAV:}prop'   => [],
-                '{DAV:}status' => 'HTTP/1.1 418 ' . \Sabre\HTTP\Response::$statusCodes[418]
-            ]);
+            if ($status != 404) { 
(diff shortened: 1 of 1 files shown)


lib/DAV/Xml/Element/Response.php
```diff
@@ -148,10 +148,12 @@ function xmlSerialize(Writer $writer) {
              * In those cases we MUST specify at least one DAV:propstat anyway, with
              * no properties.
              */
-            $writer->writeElement('{DAV:}propstat', [
-                '{DAV:}prop'   => [],
-                '{DAV:}status' => 'HTTP/1.1 418 ' . \Sabre\HTTP\Response::$statusCodes[418]
-            ]);
+            if ($status != 404) {
```
evert: This if statement appears inside of `if ($empty)`. `$empty` only is `true` when there was nothing in `$responseProperties`. `$status` will not be defined in this case, so this block will always throw an error if you hit it.

So this PR doesn't really make a lot of sense. It's not clear what the problem is, and how this would solve it.

evert:
This change makes no sense, because you are checking for the `$status` variable, in an i

PR #863, opened 2016-07-09
Removed unnecessary parameters in ClientTest

PHPGangsta:
Removed unnecessary parameters in ClientTest
The third parameter triggers "PHP warning: propPatch() expects exactly 2 parameters, 3 given"


evert:
tnx!

PR #866, opened 2016-07-09
Don't show DAV\Version::VERSION if DAV\Server::$exposeVersion = false

PHPGangsta:
The version was hidden in the browser plugin since 1.6.2. Since 2.2.0-alpha1 (commit 9a8fe641f5f2c9e9e22a00f780afc26d88defb04) it is visible again accidentally.


evert:
Makes sense, thanks!

PR #868, opened 2016-07-10
use $response in Plugin event method if possible

PHPGangsta:
fixed some PHPDoc namespaces
removed some "use namespace" which are not needed

### DIFF
--- a/lib/CalDAV/Backend/NotificationSupport.php
+++ b/lib/CalDAV/Backend/NotificationSupport.php
@@ -49,8 +49,8 @@ interface NotificationSupport extends BackendInterface {
      * If the user chose to accept the share, this method should return the
      * newly created calendar url.
      *
-     * @param string href The sharee who is replying (often a mailto: address)
-     * @param int status One of the SharingPlugin::STATUS_* constants
+     * @param string $href The sharee who is replying (often a mailto: address)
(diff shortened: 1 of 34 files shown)


evert:
Thanks! Is this lots of manual work or do you have some sort of tool to aid with this?

PHPGangsta:
It's manual work, with the help of the IDE PHPStorm. It shows errors on the right side in yellow or red. I also run specific "Inspections" to search for problems, and then went over all results.
Some things, like "unnecessary long class name", can be fixed by the IDE. Or creating an import of a long class name, and shortening all occurrences in the code. They are called "quick fixes".

There are many more things to do, like shortening class names, but I'm not sure if you have a limit. If one long class name is just used once, do you want an import of that?
Also there are many problems with "@return void". It's wrong in many places, but to fix that you also have to fix the return statements, add return statements where they are missing in some code paths.
Maybe I can create a list, where I'm not sure how to fix it, and you can have a look? Or can I mark those lines in GitHub somewhere (create a comment or so)?

Which IDE are you using?

I'm currently preparing a migration from 1.6 to 3.1, changing all our Plugins and Backends, reading code, and while doing that I'm trying to fix things... The biggest problem is our database migration with millions of table rows, which will take many hours. I will have to optimize the migration scripts to decrease the time needed for that.

PR #871, opened 2016-07-26
Fix two RFC numbers in plugin descriptions

PHPGangsta:


evert:
php-cs-fixer messed up the build again. Looks good though ;)

PR #877, opened 2016-08-25
Avoid getting 'too many variables' error with SQLite3 backend

theseer:
On a growing calendar the REPORT request gets an ever increasing list of URIs with some clients.
The current implemention will break when using SQLite3 as storage backend as it will eventually hit the build in limit for placeholders.

This simple workaround should fix that.

### DIFF
--- a/lib/CalDAV/Backend/PDO.php
+++ b/lib/CalDAV/Backend/PDO.php
@@ -512,7 +512,7 @@ SQL
         list($calendarId, $instanceId) = $calendarId;
 
         $result = [];
-        foreach(array_chunk($uris, 900) as $chunk) {
+        foreach (array_chunk($uris, 900) as $chunk) {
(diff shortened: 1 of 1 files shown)


theseer:
I'm not sure why php-cs-fixer is called for a CI test but it obviously causes the build to fail. 

That is confusing because given the patch I created, it fails on files and and formatting issues that are unrelated to my changes. All tests are still green.

evert:
Hey @theseer ,

Sorry for the CI errors. We run `php-cs-fixer` to check if pull requests comply with coding standards. The `php-cs-fixer` tool sometimes evolves and does more, and the master branch was caught behind.

I just updated master correcting this. If you merge/rebase from master the errors should go away. Thanks for the contribution!

theseer:
Done. And Fixed.

PR #879, opened 2016-08-29
Create LDAP.php

christophdb:
here is my LDAP authentification backend I am using with sabredav.

### DIFF
--- a/lib/DAV/Auth/Backend/LDAP.php
+++ b/lib/DAV/Auth/Backend/LDAP.php
@@ -1,0 +1,155 @@
+<?php
+namespace Sabre\DAV\Auth\Backend;
+
+/**
+ * LDAP authentification backend class
+ *
+ * This class can be used to authenticate for sabredav using LDAP.
+ * This function is tested with OpenLDAP on ubuntu server but should
+ * work on all other linux distributions and other LDAP implementations.
+ *
+ * @author Christoph Dyllick-Brenzinger (https://www.ionas-server.com/)
+ * @thanks to Evert Pot for this great peace of software.
+ */
+
+class LDAP extends AbstractBasic {
+
+    /**
+     * Usage of LDAP authentification
+     *
+     * assuming that this file LDAP.php is saved in the directory /vendor/sabre/dav/lib/DAV/Auth/Backend/, you can add in the server.php this new LDAP authentification backend. Add the following code:
+     * $ldap_settings = array(... your ldap settings ...);
+     * $authBackend       = new \Sabre\DAV\Auth\Backend\LDAP($ldap_settings);
+     *
+     * The LDAP authentification delivers true or false depending on the provided information.
+     * Currently there are no error messages if there is something wrong.
+     *
+     * @param array $ldap_settings
+     * @return bool
+     */
+
+
+    /**
+     * Example 1: OpenLDAP @ 127.0.0.1, plain-text communication, user in ldap tree "users",
+     * Add the following code to the server.php:
+     *
+     * $ldap_settings = array("127.0.0.1", "ou=users,dc=ionas,dc=lan", "plain");
+     * $authBackend       = new \Sabre\DAV\Auth\Backend\LDAP($ldap_settings);
+     */
+
+
+    /**
+     * Example 2: OpenLDAP @ 192.168.45.3, STARTTLS communication, user in ldap tree "users"
+     * Add the following code to the server.php:
+     *
+     * $ldap_settings = array("192.168.45.3", "ou=users,dc=example,dc=com");
+     * $authBackend       = new \Sabre\DAV\Auth\Backend\LDAP($ldap_settings);
+     */
+
+
+    /**
+     * Example 3: OpenLDAP @ 192.168.45.7:999, STARTTLS communication, custom search filter, user in ldap tree "users"
+     * Add the following code to the server.php:
+     *
+     * $ldap_settings = array(
+     *      "ldap_host"     => 192.168.45.7", 
+     *      "ldap_basedn"   => ou=users,dc=example,dc=com", 
+     *      "ldap_auth"     => starttls",
+     *      "ldap_filter"   => "(&(objectclass=inetOrgPerson)(memberOf=cn=sabredav,ou=groups,dc=ionas,dc=lan))",
+    *       "ldap_port"     => 999 );
+     * $authBackend       = new \Sabre\DAV\Auth\Backend\LDAP($ldap_settings);
+     */
+
+
+    protected $ldap;
+    function __construct($ldap) {
+        $this->ldap = $ldap;
+    }
+
+    protected function validateUserPass($username, $password) {
+
+        // transform $ldap_settings without keys to keys
+        if( $this->ldap["ldap_host"] == "" AND $this->ldap[0] != "" ){
+            $this->ldap["ldap_host"]        = $this->ldap[0];
+            $this->ldap["ldap_basedn"]      = $this->ldap[1];
+            $this->ldap["ldap_auth"]        = $this->ldap[2];
+            $this->ldap["ldap_filter"]      = $this->ldap[3];
+            $this->ldap["ldap_port"]        = $this->ldap[4];
+            
+        }
+
+        // set default values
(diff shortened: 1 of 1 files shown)


christophdb:
can anybody help me to pass the travis ci test? I have no clue why this test is still failing?!?!

evert:
Hi @christophdb ,

Instead of trying to manually make the required changes, you can simply run :

    ./bin/sabre-cs-fixer fix lib/

to automatically apply all the suggested changes.

evert:
I'm adding a list of in-line comments soon, but the main thing that is missing from this PR is unittests.

I would love to see you wrap all the `ldap_` functions you're using in a separate class. A class that would look something like :

```php

class LdapClient {

   function connect();
   function bind();
   function search();

}
```

Creating a class such as that allows you to mock it in unittest, allowing you to verify if you're correctly handling various outputs and errors.

I'm not asking you to write it if you don't feel like it, but it's important to merge the PR (and get the bounty, if you care for that). No tests = no merge.



   

}

lib/DAV/Auth/Backend/LDAP.php
```diff
@@ -0,0 +1,145 @@
...
+ * LDAP authentification backend class
+ *
+ * This class can be used to authenticate for sabredav using LDAP.
+ * This function is tested with OpenLDAP on ubuntu server but should
+ * work on all other linux distributions and other LDAP implementations.
+ *
+ * @author Christoph Dyllick-Brenzinger (https://www.ionas-server.com/)
+ * @thanks to Evert Pot for this great peace of software.
+ */
+class LDAP extends AbstractBasic {
```
evert: Call the class (and file) `Ldap` instead of `LDAP`.

I originally did not do this (I used `DAV` instead of `Dav`) and I deeply regret that. All new code should follow this pattern though. `Xml` for example does too.

lib/DAV/Auth/Backend/LDAP.php
```diff
@@ -0,0 +1,145 @@
...
+
+/**
+ * LDAP authentification backend class
+ *
+ * This class can be used to authenticate for sabredav using LDAP.
+ * This function is tested with OpenLDAP on ubuntu server but should
+ * work on all other linux distributions and other LDAP implementations.
+ *
+ * @author Christoph Dyllick-Brenzinger (https://www.ionas-server.com/)
+ * @thanks to Evert Pot for this great peace of software.
```
evert: While I much appreciate this, `@thanks` is not a standard phpdoc tag =)

Make sure that it has the following tags though:

 * @copyright Copyright (C) fruux GmbH (https://fruux.com/)
 * @license http://sabre.io/license/ Modified BSD License

lib/DAV/Auth/Backend/LDAP.php
```diff
@@ -0,0 +1,145 @@
...
+     *
+     * assuming that this file LDAP.php is saved in the directory /vendor/sabre/dav/lib/DAV/Auth/Backend/, you can add in the server.php this new LDAP authentification backend. Add the following code:
+     * $ldap_settings = array(... your ldap settings ...);
+     * $authBackend       = new \Sabre\DAV\Auth\Backend\LDAP($ldap_settings);
+     *
+     * The LDAP authentification delivers true or false depending on the provided information.
+     * Currently there are no error messages if there is something wrong.
+     *
+     * @param array $ldap_settings
+     * @return bool
```
evert: These two tags belong to the docblock for `__construct` right now they are on the docblock for `protected $ldap` which is incorrect.

lib/DAV/Auth/Backend/LDAP.php
```diff
@@ -0,0 +1,145 @@
...
+     * $ldap_settings = array(
+     *      "ldap_host"     => 192.168.45.7", 
+     *      "ldap_basedn"   => ou=users,dc=example,dc=com", 
+     *      "ldap_auth"     => starttls",
+     *      "ldap_filter"   => "(&(objectclass=inetOrgPerson)(memberOf=cn=sabredav,ou=groups,dc=ionas,dc=lan))",
+     *      "ldap_port"     => 999 );
+     * $authBackend       = new \Sabre\DAV\Auth\Backend\LDAP($ldap_settings);
+     */
+    protected $ldap;
+    function __construct($ldap) {
```
evert: I don't really think it's a good idea that the constructor has 2 different formats for setting settings. One should be enough here.

However, there is such a thing as a 'ldap uri' that, as far as I can tell, actually supports _all_ the features listed here.

Here's some more info:

* https://tools.ietf.org/html/rfc2255
* https://www.centos.org/docs/5/html/CDS/ag/8.0/LDAP_URLs-Examples_of_LDAP_URLs.html
* https://www.ldap.com/ldap-urls

You can use `Sabre\Uri\parse` to parse it. And this makes for a nice, clean, constructor.

lib/DAV/Auth/Backend/LDAP.php
```diff
@@ -0,0 +1,145 @@
...
+            $this->ldap["ldap_port"] = $this->ldap[4];
+            
+        }
+
+        // set default values
+        if ($this->ldap["ldap_port"] == "") $this->ldap["ldap_port"] = 389;
+        if ($this->ldap["ldap_auth"] == "") $this->ldap["ldap_auth"] = "starttls";
+
+        // check connection first ( http://bugs.php.net/bug.php?id=15637 )
+        $sock = @fsockopen($this->ldap["ldap_host"], $this->ldap["ldap_port"], $errno, $errstr, 1);
```
evert: Making an extra connection first is very expensive.

According to the linked ticket, both ldap_bind and ldap_start_tls *should* throw an error if the connection fails. I think that's good enough because we don't really need to know right at the beginning of `ldap_connect` whether the connection will succeed or not. We can find that out later.

lib/DAV/Auth/Backend/LDAP.php
```diff
@@ -0,0 +1,145 @@
...
+        else{
+            
+            $conn = ldap_connect($this->ldap["ldap_host"], $this->ldap["ldap_port"]);
+            ldap_set_option($conn, LDAP_OPT_PROTOCOL_VERSION, 3);
+            ldap_set_option($conn, LDAP_OPT_REFERRALS, 0);
+
+            if ($this->ldap["ldap_auth"] != "plain") {
+                if (!ldap_start_tls($conn)){
+
+                    return false;
```
evert: Throw an exception here with your specified error message.

Only return `false` is the username/password was incorrect. All connection errors should result in exceptions because then users will get some feedback around why it failed.

lib/DAV/Auth/Backend/LDAP.php
```diff
@@ -0,0 +1,145 @@
...
+            return false;
+            //return [false, "No LDAP connection. Please check the host and the port."];
+        }
+        else{
+            
+            $conn = ldap_connect($this->ldap["ldap_host"], $this->ldap["ldap_port"]);
+            ldap_set_option($conn, LDAP_OPT_PROTOCOL_VERSION, 3);
+            ldap_set_option($conn, LDAP_OPT_REFERRALS, 0);
+
+            if ($this->ldap["ldap_auth"] != "plain") {
```
evert: This if statement can be combined with the if statement on the next line.

lib/DAV/Auth/Backend/LDAP.php
```diff
@@ -0,0 +1,145 @@
...
+                if (!ldap_start_tls($conn)){
+
+                    return false;
+                    //return [false, "LDAP Connections could not be encrypted with STARTTLS. Either use ldap_auth = plain or check your STARTTLS settings"];
+
+                }
+            }
+            
+            $basedn_with_user = "cn=" . $username . "," . $this->ldap["ldap_basedn"];
+            if ($bind = @ldap_bind($conn, $basedn_with_user, $password)){
```
evert: This if statement should be reversed:

     (if (!($bind = ...)) 

So if `ldap_bind` fails, you can immediately call `return false` and you don't need an `else` statement.

DeepDiver1975:
I am closing this. Any developer an implement additional auth back-ends as needed. We deliver basic reference implementations. The rest is to to each individual project.

PR #882, opened 2016-09-08
Added button to delete files in the Browser plugin

PHPGangsta:


lib/DAV/Browser/Plugin.php
```diff
@@ -318,6 +324,12 @@ function generateDirectoryIndex($path) {
                 $buttonActions = '';
                 if ($subProps['subNode'] instanceof DAV\IFile) {
                     $buttonActions = '<a href="' . $this->escapeHTML($subProps['fullPath']) . '?sabreAction=info"><span class="oi" data-glyph="info"></span></a>';
+
+                    $buttonActions .= '<form method="post" action="" class="deleteFileForm">
+                        <input type="hidden" name="sabreAction" value="delete" />
+                        <input type="hidden" name="filePath" value="'.$this->escapeHTML($subProps['fullPath']).'" />
```
evert: Don't use the `fullPath` variable, but try to match what's already there. The path is inferred from where you are sending the `POST` request.

In this case this will break because `fullPath` is not what the server expects for sabredav servers that are not running on the root of of the domain.

lib/DAV/Server.php
```diff
@@ -1111,6 +1111,25 @@ function updateFile($uri, $data, &$etag = null) {
...
+     *
+     * This method will return true if the file was actually deleted
+     *
+     * @param string   $uri
+     * @return bool
+     */
+    function deleteFile($uri) {
+
+        if (!$this->emit('beforeUnbind', [$uri])) return false;
+
```
evert: Maybe reduce the amount of whitespace a bit in this function ;)

DeepDiver1975:
conflicting ... furthermore I question if this is needed at all .....

PR #885, opened 2016-09-14
DAV-Filesystem: Allow browsing of directorys with broken symlinks

tbsmark86:
Broken symlinks inside a Directory cause E_WARNINGs and prevent clients from viewing this directory.
This is a very simple Solution.

### DIFF
--- a/lib/DAV/FS/Directory.php
+++ b/lib/DAV/FS/Directory.php
@@ -73,7 +73,11 @@ class Directory extends Node implements DAV\ICollection, DAV\IQuota {
 
         $path = $this->path . '/' . $name;
 
-        if (!file_exists($path)) throw new DAV\Exception\NotFound('File with name ' . $path . ' could not be located');
+        if (!$this->childExists($name)) {
+
+            throw new DAV\Exception\NotFound('File with name ' . $path . ' could not be located');
+
+        }
 
         if (is_dir($path)) {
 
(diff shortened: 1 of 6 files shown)


sarnobat:
Please fix this. It makes an otherwise really good product impractical for me.

lib/DAV/FS/Directory.php
```diff
@@ -73,7 +73,11 @@ function getChild($name) {
 
         $path = $this->path . '/' . $name;
 
-        if (!file_exists($path)) throw new DAV\Exception\NotFound('File with name ' . $path . ' could not be located');
+        if (@lstat($path) === false) {
```
staabm: could we also do the fallback only when file_exists reports fall (as you did in the other cases?)

I fear we could get perf problems otherwise

staabm:
Any chance we can get a unit test for your problem?

staabm:
Since this PR is a year old a rebase would be great

tbsmark86:
I've added the UnitTest however I am quite irritated because I've could not find any for FS/File and FS/Directory.

Also, I am not sure what's the difference between FS<>FSExt? Seems to be duplicate code with some minor difference
* my new patch
* return value of File put()
* check for "." and ".." in Directory

Should I duplicate all changes to FSExt?

Side Question: Should I use FSExt for my Server is the difference notable?

evert:
The original idea behind the `FSExt` vs. `FS` is for the most part that they can be used as example, and `FS` was meant to be the 'absolute bare minimum to get a working server', `FSExt` having a lot more core features.

Not sure if that still makes sense, or if you'd want to maintain both going forward, but yea I just wanted to provide some background.

dgtlmoon:
Any chance of a merge? :)

PR #886, opened 2016-09-15
Optimize memory usage

petrkotek:
Hi!

Our company is currently running WebDAV server in production - which is powered by `sabre/dav`.

In general it works for us, however we need to be able to list directories with tens of thousands of files while having strict memory limits.

Current upstream implementation prepares all the results ("properties") into an array and then generates XML response, so memory requirements grow linearly with count of files in a directory, which doesn't really scale well. While 1000 files in a directory uses 17.3MB (peak memory usage) when listing, 50000 files uses 338 MB.

I wanted to open a discussion about this problem to discuss possible solutions.

In this PR, I implemented it ~~very simple proof of concept~~ using [Generators](http://php.net/manual/en/language.generators.php), which **reduces peak memory usage from aforementioned 338MB to 46.6MB when listing a directory with 50000 files**.

I was wondering if somebody else has come across this issue and if the repo owners or community is interested in memory optimization.

Any feedback appreciated.

Notes:
- ~~this PR is proof of concept to demonstrate how memory usage could be reduced~~
- ~tests will fail, because `Server::getPropertiesForPath()` now returns a `Generator` rather than an array; this functionality could be non-default - i.e. flip some switch (e.g. by a call of new method `Server::useGeneratorsForPropFind(true)`, which would make it non-breaking change)~~
- ~~proposed implementation optimizes only listings with depth 1~~

### DIFF
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -880,9 +880,9 @@ class Server extends EventEmitter implements LoggerAwareInterface {
      *
      * @param PropFind $propFind
      * @param array $yieldFirst
-     * @return \Generator
+     * @return \Iterator
      */
-    private function generatePathNodesRecursively(PropFind $propFind, array $yieldFirst = null) {
+    private function generatePathNodes(PropFind $propFind, array $yieldFirst = null) {
         if ($yieldFirst !== null) {
             yield $yieldFirst;
         }
@@ -909,7 +909,7 @@ class Server extends EventEmitter implements LoggerAwareInterface {
             ];
 
             if (($newDepth === self::DEPTH_INFINITY || $newDepth >= 1) && $childNode instanceof ICollection) {
-                foreach ($this->generatePathNodesRecursively($subPropFind) as $subItem) {
+                foreach ($this->generatePathNodes($subPropFind) as $subItem) {
                     yield $subItem;
                 }
             }
@@ -933,7 +933,7 @@ class Server extends EventEmitter implements LoggerAwareInterface {
      */
     function getPropertiesForPath($path, $propertyNames = [], $depth = 0) {
 
-        return iterator_to_array($this->getPropertiesGeneratorForPath($path, $propertyNames, $depth));
+        return iterator_to_array($this->getPropertiesIteratorForPath($path, $propertyNames, $depth));
 
     }
     /**
@@ -948,9 +948,9 @@ class Server extends EventEmitter implements LoggerAwareInterface {
      * @param string $path
      * @param array $propertyNames
      * @param int $depth
-     * @return \Generator
+     * @return \Iterator
      */
-    function getPropertiesGeneratorForPath($path, $propertyNames = [], $depth = 0) {
+    function getPropertiesIteratorForPath($path, $propertyNames = [], $depth = 0) {
 
         // The only two options for the depth of a propfind is 0 or 1 - as long as depth infinity is not enabled
         if (!$this->enablePropfindDepthInfinity && $depth != 0) $depth = 1;
@@ -968,7 +968,7 @@ class Server extends EventEmitter implements LoggerAwareInterface {
         ]];
 
         if (($depth > 0 || $depth === self::DEPTH_INFINITY) && $parentNode instanceof ICollection) {
-            $propFindRequests = $this->generatePathNodesRecursively(clone $propFind, current($propFindRequests));
+            $propFindRequests = $this->generatePathNodes(clone $propFind, current($propFindRequests));
         }
 
         foreach ($propFindRequests as $propFindRequest) {
@@ -1646,7 +1646,7 @@ class Server extends EventEmitter implements LoggerAwareInterface {
      *
      * If 'strip404s' is set to true, all 404 responses will be removed.
      *
-     * @param array|\Generator $fileProperties The list with nodes
+     * @param array|\Traversable $fileProperties The list with nodes
      * @param bool $strip404s
      * @return string
      */
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -1646,7 +1646,7 @@ class Server extends EventEmitter implements LoggerAwareInterface {
      *
      * If 'strip404s' is set to true, all 404 responses will be removed.
      *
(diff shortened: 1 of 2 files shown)


lib/DAV/Server.php
```diff
@@ -944,12 +971,12 @@ function getPropertiesForPath($path, $propertyNames = [], $depth = 0) {
             $parentNode
         ]];
 
-        if (($depth > 0 || $depth === self::DEPTH_INFINITY) && $parentNode instanceof ICollection) {
+        if ($depth === 1) {
+            $propFindRequests = $this->pathNodesGenerator($propFind);
```
staabm: I guess you use the generator only for `depth==1` to simplify the POC?

petrkotek: that's right. the thing is, that afaik it's not really possible to simply do recursive generator without `yield from` keyword (which is available only in PHP 7).

it should be however possible to replace the recursion in `addPathNodesRecursively()` function by an iteration

staabm: lets see what evert says. maybe we can use Generators in php7 only and use the old logic on php5. people which need high perf./low memory code would use php7 anyway.

evert:
Hi guys,

I'm definitely interested in this idea. I had this very plan for an upcoming version. I also like the idea of starting somewhere simple before expanding support to other affected areas, all the way to `getChildren` in collection nodes.

However, it would be great if this source could be refactored a bit. We definitely can do recursive generators pre PHP 7, you just need to loop through the inner generator and yield every item from there.

I would also be ok with dropping PHP 5 support entirely and just use PHP 7 and the minimum version. It might take a bit longer for that to get released though. I'd wanna make more changes such as typehints everywhere.

petrkotek:
@evert thanks for the tip with looping through the inner generator (i feel a bit stupid that i didn't realize it`:facepalm:`)

Anyways, I've rewritten the code to make it recursive. Also I changed it in a way so `getPropertiesForPath()` still returns an array (for backwards compatiblity), however internally it uses `getPropertiesGeneratorForPath()` (which obviously returns a `Generator`) - see 9b00b41.

So if _something_ is happy to get a generator (like `CorePlugin` to generate MultiStatus for `PROPFIND`), it can use the `getPropertiesGeneratorForPath()` method directly - see 3f3ae04.

Let me know, what you think.

Thanks a lot for your time.

lib/DAV/Server.php
```diff
@@ -878,12 +878,14 @@ function getHTTPHeaders($path) {
     /**
      * Small helper to support PROPFIND with DEPTH_INFINITY.
      *
-     * @param array[] $propFindRequests
      * @param PropFind $propFind
-     * @return void
+     * @param array $yieldFirst
+     * @return \Generator
```
staabm: This should be Iterator not Generator. We use a Generator to build a Iterator

lib/DAV/Server.php
```diff
@@ -929,6 +933,25 @@ private function addPathNodesRecursively(&$propFindRequests, PropFind $propFind)
      */
     function getPropertiesForPath($path, $propertyNames = [], $depth = 0) {
 
+        return iterator_to_array($this->getPropertiesGeneratorForPath($path, $propertyNames, $depth));
```
staabm: 👍

lib/DAV/Server.php
```diff
@@ -878,12 +878,14 @@ function getHTTPHeaders($path) {
...
      *
-     * @param array[] $propFindRequests
      * @param PropFind $propFind
-     * @return void
+     * @param array $yieldFirst
+     * @return \Generator
      */
-    private function addPathNodesRecursively(&$propFindRequests, PropFind $propFind) {
-
+    private function generatePathNodesRecursively(PropFind $propFind, array $yieldFirst = null) {
```
staabm: I would shorten the name to `generatePathNodes()`

lib/DAV/Server.php
```diff
@@ -929,6 +933,25 @@ private function addPathNodesRecursively(&$propFindRequests, PropFind $propFind)
...
+     * 'allprops' is assumed.
+     *
+     * If a depth of 1 is requested child elements will also be returned.
+     *
+     * @param string $path
+     * @param array $propertyNames
+     * @param int $depth
+     * @return \Generator
+     */
+    function getPropertiesGeneratorForPath($path, $propertyNames = [], $depth = 0) {
```
staabm: Generator -> Iterator

lib/DAV/Server.php
```diff
@@ -1631,9 +1650,14 @@ function getResourceTypeForNode(INode $node) {
      * @param bool $strip404s
      * @return string
      */
-    function generateMultiStatus(array $fileProperties, $strip404s = false) {
+    function generateMultiStatus($fileProperties, $strip404s = false) {
```
staabm: We need to update the phpdoc to `array|Traversable` IMO

lib/DAV/Server.php
```diff
@@ -929,6 +933,25 @@ private function addPathNodesRecursively(&$propFindRequests, PropFind $propFind)
      */
     function getPropertiesForPath($path, $propertyNames = [], $depth = 0) {
```
staabm: Should we deprecate this method and mention the new Iterator based method call the successor? Or do you think there are still good usecases in the future in which a iterator is not sufficient?

petrkotek: Good point, I think there shouldn't be a need for array-based method anymore.

 I'm happy to add `@deprecated` annotation to it, however I guess it would be nice to get @evert's feedback too.

If you want to rewrite usages of `getPropertiesForPath` to `getPropertiesIteratorForPath()`, I'd prefer to do it in separate PR since it will also involve lots of unit test changes.

staabm: 👍

petrkotek: I've added the deprecation message - see commit 3746382.

evert: agreed with adding @deprecated

evert:
Ready to merge from your perspective @petrkotek ?

petrkotek:
@evert: yep, i'm happy with it how it stands. happy to get it merged ⭐

PR #887, opened 2016-09-19
Tests cleanup

petrkotek:
This is low priority PR, but also shouldn't be difficult to review since it changes only tests (apart from one-line commit which affects actual code).

When troubleshooting certain issues, I had to go through some unit tests and couldn't help myself to do a bit of:
- code-style cleanup
- PHPDoc improvements
- removal of unnecessary `require_once` statements
- removal of `ResponseMock` class
- minor tests refactoring.


lib/DAVACL/Plugin.php
```diff
@@ -533,7 +533,6 @@ function getSupportedPrivilegeSet($node) {
             'aggregates' => $this->getSupportedPrivilegeSet($node)
         ];
 
-        $fpsTraverse = null;
```
evert: I don't think this is correct. The $fpsTraverse needs to be passed as a reference, and for that to happen the variable needs to exist.

I think (but I'm not sure) that this would trigger an `E_NOTICE`.

tests/Sabre/HTTP/ResponseMock.php
```diff
@@ -1,22 +0,0 @@
...
-/**
- * HTTP Response Mock object
- *
- * This class exists to make the transition to sabre/http easier.
- *
- * @copyright Copyright (C) fruux GmbH (https://fruux.com/)
- * @author Evert Pot (http://evertpot.com/)
- * @license http://sabre.io/license/ Modified BSD License
- */
-class ResponseMock extends Response {
```
evert: YAY

evert:
#892 replaces this PR

PR #888, opened 2016-09-20
Optimize memory usage: implement "ArrayCache" and limit cache size

petrkotek:
I'd like to propose one more memory usage optimization.

Issue this PR is solving is that cache size in `Sabre\DAV\Tree` (`$cache`) is currently *unlimited*.

When count of items grows to tens of thousands, size of `$cache` array grows linearly and can end up exhausting all the available memory.

For example in our case, when WebDAV collection listings contains 50000 items, cache uses around ~22MB (about 440B per item). When there is even more items in the collection, memory gets exhausted.

This PR adds `Cache` interface and simple implementation - `ArrayCache`, which has a cap on count of items in the cache -- it simply stops storing new items in the cache when cache capacity is reached.

Open questions:
- Q1: When the cache capacity is reached, would it be better to evict some of the existing items? would be "random" good enough eviction strategy?
- Q2: Is it fine to create new `Sabre\Common` namespace (just for the `Cache` and `ArrayCache` classes) or is there a better place for these classes?

### DIFF
--- a/lib/Common/ArrayCache.php
+++ b/lib/Common/ArrayCache.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace Sabre\Common;
+
+use ArrayIterator;
+
+/**
+ * Implementation of Cache interface using a simple associative array.
+ */
+class ArrayCache implements Cache {
+
+    const DEFAULT_CACHE_SIZE = 1000;
+
+    /** @var int Maximum count of items in the cache */
+    private $cacheSize;
+
+    /** @var mixed[] */
+    private $cache = [];
+
+    /**
+     * @param int $cacheSize
+     */
+    function __construct($cacheSize = self::DEFAULT_CACHE_SIZE) {
+
+        $this->cacheSize = $cacheSize;
+
+    }
+
+    /**
+     * Retrieves an item from the cache.
+     *
+     * @param string $key
+     * @return mixed|null
+     */
+    function get($key) {
+
+        return array_key_exists($key, $this->cache) ? $this->cache[$key] : null;
+
+    }
+
+    /**
+     * Saves an item to the cache.
+     *
+     * When the capacity is reached, no new items will be added. Existing items under will still get updated even if
+     * the capacity is reached.
+     *
+     * @param string $key
+     * @param mixed $value
+     * @return void
+     */
+    function set($key, $value) {
+
+        $addingNewItem = !array_key_exists($key, $this->cache);
+        if ($addingNewItem && count($this->cache) >= $this->cacheSize) {
+            // don't store new values, if the cache is full
+            return;
+        }
+        $this->cache[$key] = $value;
+
+    }
+
+    /**
+     * Removes an item from the cache.
+     *
+     * @param string $key
+     * @return void
+     */
+    function remove($key) {
+
+        unset($this->cache[$key]);
+
+    }
+
+    /**
+     * Checks whether an item is in the cache.
+     *
+     * @param string $key
+     * @return bool
+     */
+    function keyExists($key) {
+
+        return array_key_exists($key, $this->cache);
+
+    }
+
+    /**
+     * Returns all items in the cache (key-value pairs).
+     *
+     * @return \Traversable
+     */
+    function getAll() {
+
+        return new ArrayIterator($this->cache);
+
+    }
+
+}
--- a/lib/DAV/Tree.php
+++ b/lib/DAV/Tree.php
@@ -2,6 +2,8 @@
 
 namespace Sabre\DAV;
 
+use Sabre\Common\ArrayCache;
+use Sabre\Common\Cache;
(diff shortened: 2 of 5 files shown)


lib/Common/ArrayCache.php
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * Retrieves an item from the cache.
+     *
+     * @param string $key
+     * @return mixed|null
+     */
+    function get($key) {
+
+        return array_key_exists($key, $this->cache) ? $this->cache[$key] : null;
```
staabm: you should use a combination of isset & array_key_exists.

```
if (isset($this->cache[$key])) return $this->cache[$key];
return array_key_exists($key, $this->cache) ? $this->cache[$key] : null;
```

that way you combine the way faster isset (for positive cache hits) but preserve the ability to store `null` as a value

lib/DAV/Tree.php
```diff
@@ -27,20 +29,22 @@ class Tree {
      * This is the node cache. Accessed nodes are stored here.
      * Arrays keys are path names, values are the actual nodes.
```
staabm: the docs should be updated: `Arrays keys` -> `Cache keys`

lib/Common/ArrayCache.php
```diff
@@ -0,0 +1,97 @@
...
+namespace Sabre\Common;
+
+use ArrayIterator;
+
+/**
+ * Implementation of Cache interface using a simple associative array.
+ */
+class ArrayCache implements Cache {
+
+    const DEFAULT_CACHE_SIZE = 1000;
```
staabm: I have no experience whether this is a "good upper limit"

lib/Common/ArrayCache.php
staabm: same in the other methods of this class

staabm:
> For example in our case, when WebDAV collection listings contains 50000 items, cache uses around ~22MB (about 440B per item). When there is even more items in the collection, memory gets exhausted.

22MB is not that much of a problem, is it?

evert:
I'm a little unsure about the usefulness of this. In particular:

> For example in our case, when WebDAV collection listings contains 50000 items, cache uses around ~22MB (about 440B per item). When there is even more items in the collection, memory gets exhausted.

22MB does not really sound like a lot to me. Especially given these massive request are probably a little rare. If a the odd request spikes up 22 megs from time to time, this wouldn't really be a reason for concern to me.

I'm tempted to say, I don't want the extra complexity given that the benefit seems pretty minimal. Plus for special cases you would actually be able to swap out your `Tree` object for another that is more specifically optimized for your use-case. So you're not entirely stuck, even if you decide you do need this.

Thoughts?

petrkotek:
yeah, 22MB shouldn't be a problem in most of the cases, but sometimes cached data is larger (e.g. long paths) and sometimes there is not 50k items, but 150k+ items and then it becomes a problem (with our 128MB memory limit)

issue i see is that the cache can grow "infinitely".

if you're not interested in this optimization, we really can swap out the `Tree` implementation as you mentioned, that's not a problem (although it means we'll need to maintain it as interface may slightly change as new versions of `sabre/dav` get released)

(I understand that it looks like lots of code solving relatively small issue, however i believe in this case it's good practice to separate caching concerns out of the `Tree` object)

staabm:
> yeah, 22MB shouldn't be a problem in most of the cases, but sometimes cached data is larger (e.g. long paths) and sometimes there is not 50k items, but 150k+ items and then it becomes a problem (with our 128MB memory limit)

did you experience a case where it was a problem and this patch solved this issue?

petrkotek:
Yes, we're running our (private) fork of `sabre/dav-1.8.x`, where we needed first to do the "iterators/generators fix" and then after couple of months we also needed to do this "capped cache fix" as it turned out that some of our clients have rather insane count of files in some directories ("collections"). In other words, if memory requirements grow with count of items in a collection, we'll get "memory exhausted" fatal error sooner or later :)

(Now we're upgrading to the latest version of `sabre/dav` to get PHP7 compatibility)

evert:
I can see that it becomes a problem at 150K items, but I think I would be tempted to try and figure out ways to reduce the total number, instead of supporting it.

But even then, based on your own numbers 150K items would result in 66MB of memory. To me that's acceptable if those requests happen rarely. I also think it's fine to set the PHP memory limit to something like 1G. You definitely don't want a large portion of requests to consume large amount of memory, but as long as you know it's a small subset this should be acceptable.

What I see in this patch is an increase in complexity (and CPU time) that's a negative for 99.9% of use-cases, and an optimization for 0.1%. 

If we can eventually switch the entire engine to be generator based, we might be able to do away entirely with caching though. The request you're talking about is probably a `PROPFIND`, and it would be nice if we can just deal with every node one-by-one without storing them for potential later use.

PR #890, opened 2016-09-28
Memory optimization: support responses bigger than max allowed memory

petrkotek:
## What?
Use temporary file for multi-status responses with 10000+ entries (& flush every 10000 entries).

## Why?
`$w->outputMemory()` doesn't work well for large responses (around 10MB, but may depend on php settings). ⚠️ Truncated response may be returned. ⚠️ 

Note 1: existing code works perfectly fine for responses <10MB)
Note 2: code inspired by http://codeinthehole.com/writing/creating-large-xml-files-with-php/

### DIFF
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -1651,7 +1651,7 @@ class Server extends EventEmitter implements LoggerAwareInterface {
      *
      * @param array|\Traversable $fileProperties The list with nodes
      * @param bool $strip404s
-     * @return string
+     * @return string|resource String for less than 10000 entries. For more entries, returns a resource (file handler).
      */
     function generateMultiStatus($fileProperties, $strip404s = false) {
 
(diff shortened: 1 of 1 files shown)


lib/DAV/Server.php
```diff
@@ -1651,7 +1651,7 @@ function getResourceTypeForNode(INode $node) {
      *
      * @param array|\Traversable $fileProperties The list with nodes
      * @param bool $strip404s
-     * @return string
+     * @return string|resource String for less than 10000 entries. For more entries, returns a resource (file handler).
```
petrkotek: it's safe to return `resource` as result of `generateMultiStatus()` is always used as an argument to `$response->setBody()`, which accepts both `string` & `resource`.

evert:
Instead of this approach, I would actually really like to refactor the sabre/http library a bit to allow callbacks to be specified as setBody(). Outputting the body would then call the function that was passed, which is then responsible for echoing to STDOUT.

This avoids the need for a temporary file, and makes this truly streaming instead of buffered. So for that reason I'd reject this, as I think this is 2 step forwards and 1 step back ;)

petrkotek:
Sounds like a bit more work, however cleaner & more efficient indeed. Let me have a look in that.

petrkotek:
groundwork in `sabre/http` to support callbacks as a body: https://github.com/fruux/sabre-http/pull/65

evert:
Closing this as the sabre/http approach is under way.

Issue #322, opened 2013-03-17
Authentication against IMAP server


rosali:
I'd really love to see a method to authenticate against an IMAP server.

We recommend to use SabreDAV along with our MyRoundcube plugins (http://myroundcube.com/myroundcube-plugins/sabredav-plugin).

It would be very handy to such an authentication backend when integrating SabreDAV into a Webmail suite.

Biniou180:
Hello!

I know this issue is old, however I had the same problem, and the example above is broken.
As this page often comes in the firsts results of search engines, I post my solution for people who might want to connect their dovecot to sabredav ;-)

In the example of Steffenweber, there was an error concerning the mt_rand function, and as exceptions didn't raise their causes, it was not very clear.

I have juste changed 0xffffffff`` by `mt_getrandmax()` and raised messages when exceptions occurs. `0xffffffff` was returning -1 and `mt_rand(0, -1)` raises an error.

My corrected configuration (`dovecot-sasl.php` file):
```php
<?php
/**
 * http://wiki.dovecot.org/Authentication%20Protocol
 */
class DovecotAuth
{
    private $fp;
    private $service;
    private $mechanisms = [];

    const VERSION_MAJOR = 1;
    const VERSION_MINOR = 1;
    const DEBUG = false;

    public function __construct($fp, $service)
    {
        $this->fp = $fp;
        $this->service = $service;
    }

    public function authenticate($username, $password)
    {
        $pid = getmypid();
        $id = mt_rand(0, mt_getrandmax());
        $data = base64_encode("$username\0$username\0$password");

        $this->send(['VERSION', self::VERSION_MAJOR, self::VERSION_MINOR]);
        $this->send(['CPID', $pid]);
        $this->recv(['DONE']);

        if (!in_array('PLAIN', $this->mechanisms))
        {
            throw new \Exception('SASL: server does not support PLAIN mechanism');
        }

        $this->send(['AUTH', $id, 'PLAIN', 'service=' . $this->service]);
        $this->recv(['CONT']);

        $this->send(['CONT', $id, $data]);
        $status = $this->recv(['OK', 'FAIL', 'CONT']);

        fclose($this->fp);
        return $status === 'OK';
    }

    private function send(array $data)
    {
        $str = implode("\t", $data) . "\n";

        if (self::DEBUG) echo "Send: $str";
        fwrite($this->fp, $str);
    }

    private function recv(array $needles)
    {
        while ($line = fgets($this->fp))
        {
[...]

PR #891, opened 2016-09-28
Add IMAP authentication

c0d3z3r0:
This add IMAP authentication.
Work in progress! It's working already but I need to add a php IMAP server for testing...

### DIFF
--- a/lib/DAV/Auth/Backend/IMAP.php
+++ b/lib/DAV/Auth/Backend/IMAP.php
@@ -1,0 +1,90 @@
+<?php
+
+namespace Sabre\DAV\Auth\Backend;
+
+/**
+ * This is an authentication backend that uses imap.
+ *
+ * @copyright Copyright (C) fruux GmbH (https://fruux.com/)
+ * @author Michael Niewöhner (foss@mniewoehner.de)
+ * @author rosali (https://github.com/rosali)
+ * @author Evert Pot (http://evertpot.com/)
+ * @license http://sabre.io/license/ Modified BSD License
+ */
+class IMAP extends AbstractBasic {
+
+    /**
+     * IMAP server in the form host[:port]
+     *
+     * @var string
+     */
+    protected $imap_host;
+
+    /**
+     * IMAP connection flags
+     * @see http://php.net/manual/en/function.imap-open.php
+     *
+     * @var string
+     */
+    protected $imap_flags;
+
+    /**
+     * Creates the backend object.
+     * imap_host is the IMAP server in the form host[:port]
+     * imap_flags specifies the IMAP connection flags
+     * @see http://php.net/manual/en/function.imap-open.php
+     *
+     * @param string $imap_host
+     * @param string $imap_flags
+     */
+    function __construct($imap_host, $imap_flags = '') {
+
+        $this->imap_host = $imap_host;
+        $this->imap_flags = $imap_flags;
+
+    }
+
+    /**
+     * Connects to an IMAP server and tries to authenticate
+     *
+     * @param string $username
+     * @param string $password
+     * @return bool
+     */
+    protected function imapOpen($username, $password) {
+
+        $mailbox = "{" . $this->imap_host . $this->imap_flags . "/readonly}";
+
+        $success = false;
+
+        try {
+            $imap = imap_open($mailbox, $username, $password, 0, 1);
+            if ($imap)
+                $success = true;
+        } catch (\ErrorException $e) {
+            error_log($e->getMessage());
+        }
+
+        foreach (imap_errors() as $error)
+            error_log($error);
+
+        if (isset($imap) && $imap)
+            imap_close($imap);
+
+        return $success;
+    }
+
+    /**
+     * Validates a username and password by trying to authenticate against IMAP
+     *
+     * @param string $username
+     * @param string $password
+     * @return bool
+     */
+    protected function validateUserPass($username, $password) {
+
+        return $this->imapOpen($username, $password);
+
+    }
+
+}
--- a/tests/Sabre/DAV/Auth/Backend/IMAPTest.php
+++ b/tests/Sabre/DAV/Auth/Backend/IMAPTest.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Sabre\DAV\Auth\Backend;
+
+class IMAPTest extends \PHPUnit_Framework_TestCase {
+
+    function testGoodPassword() {
+
+        $imap_host = 'localhost:9993';
+        $imap_flags = '';
+        $imap = new IMAPMock($imap_host, $imap_flags);
+        $this->assertTrue($imap->validateUserPass('username', 'password'));
+
+    }
+
+    function testBadPassword() {
+
+        $imap_host = 'localhost:9993';
+        $imap_flags = '';
+        $imap = new IMAPMock($imap_host, $imap_flags);
+        $this->assertFalse($imap->validateUserPass('username', 'badpassword'));
+
+    }
+
+}
+
+
+class IMAPMock extends IMAP {
+
+    /**
+     * Connects to an IMAP server and tries to authenticate
+     *
+     * @param string $username
+     * @param string $password
+     * @return bool
+     */
+    function imapOpen($username, $password) {
+
+        return ($username == 'username' && $password == 'password');
+
+    }
+
+    /**
+     * Validates a username and password
+     *
+     * This method should return true or false depending on if login
+     * succeeded.
+     *
+     * @param string $username
+     * @param string $password
+     * @return bool
+     */
+    function validateUserPass($username, $password) {
+
+        return parent::validateUserPass($username, $password);
+
+    }
+
+}


evert:
Hi @c0d3z3r0 ,

Cool to see that you are picking this up. I'm ok with _not_ having a real IMAP server in the unittests, however I really would like to see that you create a small class that wraps the imap_ functions so you can mock it from inside PHPUnit.

Since you are going for the bounty, it's _required_ for this feature.

Let me know if you have any questions! I'm excited

bakerjalexander:
I've got a real world server I can test it with.  Running Dovecot.

Hitman86R:
Also I have a world server that runs an imap with dovecot. tonight I try

c0d3z3r0:
@bakerjalexander @Hitman86R I've already tested it on my mailserver where it worked. Please let me know if there are any problems on yours

c0d3z3r0:
Okay, I think this looks good now.
@evert what do you say?

c0d3z3r0:
Added documentation: https://github.com/fruux/sabre.io/pull/96

Hitman86R:
Now I am at work, when I get back home I try it.

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -0,0 +1,79 @@
...
+ * @license http://sabre.io/license/ Modified BSD License
+ */
+class IMAP extends AbstractBasic {
+
+    /**
+     * IMAP server and options
+     *
+     * @var string
+     */
+    protected $imap;
```
evert: I don't think this property is actually used?

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Creates the backend object.
+     *
+     * @param string $imap
+     */
+    function __construct($imap_host, $imap_flags = '') {
+
+        $this->imap_host = $imap_host;
+        $this->imap_flags = $imap_flags;
```
evert: These properties are undefined

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -0,0 +1,79 @@
...
+        try {
+            $imap = imap_open($mailbox, $username, $password, 0, 1);
+
+            if ($imap) {
+                imap_close($imap);
+                $success = true;
+            } else {
+              $success = false;
+            }
+        } catch (\Exception $e) {
```
evert: What kind of exceptions can be emitted from this? If there's a real connection error I think it would be nice if that exception still gets emitted to the user, so they have a chance to debug this.

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -0,0 +1,79 @@
...
+            $success = false;
+        }
+
+        foreach (imap_errors() as $error)
+            error_log($error);
```
evert: This should probably be an exception

lib/DAV/Auth/Backend/IMAP.php
c0d3z3r0: Nope. I forgot to change that. fixed

lib/DAV/Auth/Backend/IMAP.php
c0d3z3r0: Yep. Fixed with issue above

lib/DAV/Auth/Backend/IMAP.php
c0d3z3r0: imap_open treats some warnings as errors for example when using an unencrypted connection. This also happens when you're connecting locally through loopback where you should not need any encryption (if anyone has access to loopback... well... you know...).

The problem with this is that when you don't clear the errors as above imap_connect will emmit an exception. That's not what we want when it is "only" a warning.

I wasn't sure how to solve this in a clean way - ideas welcome :-)

staabm: Do those errors have error codes, a indicator for severity or something like that?

c0d3z3r0: Okt 10 21:38:24 c0xu4.mniewoehner.de sabre-dav[597]: PHP Fatal error:  Uncaught exception 'ErrorException' with message 'Unknown: SECURITY PROBLEM: insecure server advertised AUTH=PLAIN (errflg=1)' in Unknown:0
                                                     Stack trace:
                                                     #0 [internal function]: exception_error_handler(8, 'Unknown: SECURI...', 'Unknown', 0, Array)
                                                     #1 {main}
                                                       thrown in Unknown on line 0

lib/DAV/Auth/Backend/IMAP.php
c0d3z3r0: yeah, I agree. fixed

lib/DAV/Auth/Backend/IMAP.php
evert: Hmm... bit hard to tell. That definitely seems like the type of thing you would want to be able to surpress and not always get a fatal error.

So lets keep this the way it was then. Logging them is nice.

lib/DAV/Auth/Backend/IMAP.php
c0d3z3r0: I've tested this again and always get that ErrorException so it seems we need it...

evert: This might be due to calling `imap_close` _before_ `imap_errors()`. 

However, if that doesn't fix it you should catch that specific exception (ErrorException).

c0d3z3r0: No, I've already changed that on my test machine... I'll readd the catch block

c0d3z3r0: wow strange...  now I know why I get this.

When the username/password is wrong, an ErrorException happens saying "Couldn't open stream". Because of that exception we don't get to imap_errors and the ErrorException caused by the "security problem" is printed.

It seems we need the try-catch block to catch a failed authentication. (The documentation says, imap_open returns false when there where errors. I have no clue why it throws an exception o.O)

c0d3z3r0: I checked that, but no. I get the feeling this is a sabre problem... when I test this script as cli I don't get an ErrorException but a warning. Then $imap is false like the documentation says.

Is sabre-dav treating php warnings as error exceptions?

<?php

$mailbox = "{localhost:993/notls/norsh/readonly}";
$success = false;

try {

  $imap = imap_open($mailbox, "bar", "foo", 0, 1);
  print("imap\n"); 
  if ($imap)
    $success = true;

} catch (\ErrorException $e) {
  $success = false;
  print("ex\n");
  print($e . "\n");
}

foreach (imap_errors() as $error)
  print($error . "\n");

if(defined($imap) && $imap)
  imap_close($imap);

if ($success) print "ok\n"; else print "nok\n";


Output:
PHP Warning:  imap_open(): Couldn't open stream {localhost:993/notls/norsh/readonly} in /tmp/x.php on line 8
imap
SECURITY PROBLEM: insecure server advertised AUTH=PLAIN
Can not authenticate to IMAP server: [AUTHENTICATIONFAILED] Authentication failed.
nok

evert: Correct. By default any notice, warning or error gets converted to an ErrorException. This is definitely by design. If you specifically want to supress these you can use the `@` operator on imap_open, but ideally this is avoided. If there's no way around it, that's what needs to happen though.

c0d3z3r0: Oh. Just saw that, too ;-)
"* Mapping PHP errors to exceptions." .....

I don't like the idea of suppressing errors with the @ op. Catching the exception and logging it is the best solution I think

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -0,0 +1,76 @@
...
+        $imap = imap_open($mailbox, $username, $password, 0, 1);
+
+        if ($imap) {
+            imap_close($imap);
+            $success = true;
+        } else {
+            $success = false;
+        }
+
+        foreach (imap_errors() as $error)
```
c0d3z3r0: done

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -0,0 +1,79 @@
...
+ */
+class IMAP extends AbstractBasic {
+
+    /**
+     * IMAP server and options
+     *
+     * @var string
+     */
+    protected $imap_host;
+    protected $imap_flags;
```
evert: This needs its own docblock.

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Creates the backend object.
+     *
+     * @param string $imap
+     */
+    function __construct($imap_host, $imap_flags = '') {
```
evert: $imap_flags is not really documented. It would be nice if it were in the docblock, but it would be even better if the constructor has a paragraph describing what it is. If this simply means you just need to add a link to the PHP documentation, then this is good enough ;)

tests/Sabre/DAV/Auth/Backend/IMAPTest.php
```diff
@@ -0,0 +1,59 @@
...
+     * Validates a username and password
+     *
+     * This method should return true or false depending on if login
+     * succeeded.
+     *
+     * @param string $username
+     * @param string $password
+     * @return bool
+     */
+    function validateUserPass($username, $password) {
```
evert: If this function just inherits everything from it's parent, just completely remove it!

c0d3z3r0: validateUserPass in class IMAP is *protected* so it can't be called from the "outside"

lib/DAV/Auth/Backend/IMAP.php
c0d3z3r0: done

tests/Sabre/DAV/Auth/Backend/IMAPTest.php
evert: Makes sense, I missed that.

evert:
Thank you, this is great for a first version. You can claim the bounty on bountysource now if you'd like!

c0d3z3r0:
:+1: #322 can be closed now

c0d3z3r0:
@evert Looks like you need to close the issue on bountysource. It says the bug is still open

sebastian-schmid:
Hello @evert @c0d3z3r0 ,

I tested this piece of code but on my first try it did not work.

I think in IMAP.php the following part has to be part of the catch block:
`foreach (imap_errors() as $error)`
`  error_log($error);`

after I moved the code into the catch block it works. 

Better than a /readonly connect to IMAP is the OP_HALFOPEN option I think, should be faster and no mailbox will be touched:
`$imap = imap_open($mailbox, $username, $password, OP_HALFOPEN, 1);`

c0d3z3r0:
@sebastian-schmid the imap_errors have to be read in _every_ case. Otherwise you'd get an Exception / PHP Notice: https://secure.php.net/manual/de/function.imap-errors.php

What do you mean by "it did not work"? Were there any errors, exceptions etc.?

OP_HALFOPEN may be a good idea but needs to be tested.

sebastian-schmid:
@c0d3z3r0 yes, it  did not work - means authentication did not work. IMAP open was successful as I could see in the dovecot log. But on calendar clients I got "invalid User and/or password".

And over Webbrowser I got "Invalid argument supplied for foreach()" error message in sabre/dav xml response. After moving this into the catch block it worked as expected.

OP_HALFOPEN is working at my installation, but would be great if somebody else can verify this.

c0d3z3r0:
I have to check this on my setup....
In the meantime, could you provide some more information (PHP version etc.), logs from php-fpm, the exact xml response...?

c0d3z3r0:
@sebastian-schmid IMAP auth is working correctly on my systems... also the travis checks passed without any errors. I have no clue what's wrong :/

sebastian-schmid:
Of course:

PHP Version: 5.5.38 (PHP nativ no php-fpm)
Dovecot: 2.1.7

sabre/dav XML response:
<d:error>
<s:sabredav-version>3.2.0</s:sabredav-version>
<s:exception>ErrorException</s:exception>
<s:message>Invalid argument supplied for foreach()</s:message>
</d:error>

In Apache Logs are no errors logged and dovecot login is successful.

c0d3z3r0:
I've opened an issue for that... #915

Hitman86R:
Hi,
sorry if I didn't answer this discussion anymore, but I had big family problems.
returning to us in these days I'm testing the login script via imap, unfortunately without success. Going to examine the system syslog I noticed that the login data is not passed to dovecot, going into connection timeout. understand also that I have not followed the development of the situation, so if someone already good heart can help me.

PR #892, opened 2016-10-10
Petrkotek tests cleanup

evert:
This PR is like #887, but reverts one change. Thanks a lot @petrkotek !


petrkotek:
I'm glad I could contribute 👍

PR #894, opened 2016-10-10
Using PHP7 strict_types everywhere in the source

evert:


lib/CalDAV/SharingPlugin.php
```diff
@@ -304,6 +304,7 @@ function httpPost(RequestInterface $request, ResponseInterface $response) {
 
                 if ($url) {
                     $writer = $this->server->xml->getWriter();
+                    $writer->contextUri = $request->getUrl();
```
staabm: Unrelated change?

evert: No, this is initialized as `null` and caused some issues further down the line.

PR #898, opened 2016-10-12
Sabre\DAV\Server: support large PROPFINDs

petrkotek:
Support large responses by returning a callback generating the response for `PROPFIND` rather than generating a string in memory.

(Callbacks are is now supported since https://github.com/fruux/sabre-http/pull/65 is now merged)

### DIFF
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -1643,42 +1643,43 @@ class Server implements LoggerAwareInterface, EmitterInterface {
 
 
     /**
-     * Generates a WebDAV propfind response body based on a list of nodes.
+     * Returns a callback generating a WebDAV propfind response body based on a list of nodes.
      *
      * If 'strip404s' is set to true, all 404 responses will be removed.
      *
      * @param array|\Traversable $fileProperties The list with nodes
      * @param bool $strip404s
-     * @return string
+     * @return callable
      */
     function generateMultiStatus($fileProperties, $strip404s = false) {
 
         $w = $this->xml->getWriter();
-        $w->openMemory();
-        $w->contextUri = $this->baseUri;
-        $w->startDocument();
+        return function() use ($fileProperties, $strip404s, $w) {
+            $w->openUri('php://output');
+            $w->contextUri = $this->baseUri;
+            $w->startDocument();
 
-        $w->startElement('{DAV:}multistatus');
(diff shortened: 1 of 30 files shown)


staabm:
Should we add this in a BC way or is this mewthod considered internalApi ?

staabm:
How large is large? How big was possible before and how much does it improve with this change?

petrkotek:
"large" depends on PHP settings (mostly memory limit) and size of the response (number of records & length of individual records) as current approach needs to be able to generate whole response in memory.

In our case "large" is somewhere around 30000 items. With this fix in place, memory requirements shouldn't grow with result set size, so in theory no limit (in practice there could be some memory leak (?))

petrkotek:
ad backwards compatibility: since result of `generateMultiStatus()` is always used as an arguemnt for `$response->setBody()` and new `sabre/http`'s response now supports callbacks as "body" (https://github.com/fruux/sabre-http/pull/65), it should be safe (i just had to update the tests as they were assuming that th`$response->body` is `string|stream`)

rayward:
@staabm is this something we can move forward with? we're relying on an internal fork right now with this change, it'd be nice to get back onto the canonical.

evert:
Given that this is a public API, I think it's a BC break and should probably only make it into a new BC-breaking version.

rayward:
Thats fair, are we able to target the next major version then?

As mentioned in the top comment, this has supporting changes in master for sabre/http (https://github.com/sabre-io/http/pull/65) which is targeting 5.0 of that component.

DeepDiver1975:
Long long time passed by .... what a pity ....
@petrkotek would you still be available to bring this back into shape?

DeepDiver1975:
A more general question on this: we don't really know the content length before we are finished writing the full content. So basically we cannot send a content length header.
- [ ] Do we know if the broad mass of clients can live with this?
- [ ] Do we need to switch chunked encoding? (I only did a quick research on this ... maybe I'm on the wrong track)
- [ ] we might want to add an option to still have the old behavior for compat reasons

evert:
PHP will switch to chunked encoding automatically if there's no content-length header.

DeepDiver1975:
We have this now in place after merging #1204 

Thanks @petrkotek for the initial work! :metal:

PR #914, opened 2016-11-26
Rework Auth\IMAP

c0d3z3r0:
- use option OP_READONLY instead of appending '/readonly'
- use OP_HALFOPEN to speed up authentication
- adapt test

### DIFF
--- a/lib/DAV/Auth/Backend/IMAP.php
+++ b/lib/DAV/Auth/Backend/IMAP.php
@@ -51,8 +51,10 @@ class IMAP extends AbstractBasic {
             error_log($e->getMessage());
         }
 
-        foreach (imap_errors() as $error)
-            error_log($error);
+        $errors = imap_errors();
+        if ($errors)
+            foreach ($errors as $error)
+                error_log($error);
 
         if (isset($imap) && $imap)
             imap_close($imap);
--- a/lib/DAV/Auth/Backend/IMAP.php
+++ b/lib/DAV/Auth/Backend/IMAP.php
@@ -44,7 +44,7 @@ class IMAP extends AbstractBasic {
         $success = false;
 
         try {
-            $imap = imap_open($this->mailbox, $username, $password, OP_HALFOPEN|OP_READONLY, 1);
+            $imap = imap_open($this->mailbox, $username, $password, OP_HALFOPEN | OP_READONLY, 1);
             if ($imap)
                 $success = true;
         } catch (\ErrorException $e) {
--- a/lib/DAV/Auth/Backend/IMAP.php
+++ b/lib/DAV/Auth/Backend/IMAP.php
@@ -14,33 +14,21 @@ namespace Sabre\DAV\Auth\Backend;
 class IMAP extends AbstractBasic {
 
     /**
-     * IMAP server in the form host[:port]
-     *
-     * @var string
-     */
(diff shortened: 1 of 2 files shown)


c0d3z3r0:
#891

c0d3z3r0:
#915

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -53,25 +45,26 @@ function __construct($imap_host, $imap_flags = '') {
...
+            $imap = imap_open($this->mailbox, $username, $password, OP_HALFOPEN | OP_READONLY, 1);
             if ($imap)
                 $success = true;
         } catch (\ErrorException $e) {
             error_log($e->getMessage());
         }
 
-        foreach (imap_errors() as $error)
-            error_log($error);
+        $errors = imap_errors();
```
staabm: 👍

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -37,10 +30,9 @@ class IMAP extends AbstractBasic {
      * @param string $imap_host
      * @param string $imap_flags
      */
-    function __construct($imap_host, $imap_flags = '') {
+    function __construct($mailbox) {
```
staabm: In case this code was ever part of a regular release, we cannot change this that way, because BC

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -53,25 +45,26 @@ function __construct($imap_host, $imap_flags = '') {
...
+        $errors = imap_errors();
+        if ($errors)
+            foreach ($errors as $error)
+                error_log($error);
 
         if (isset($imap) && $imap)
             imap_close($imap);
 
         return $success;
+
```
staabm: Unintnded whitespace change?

lib/DAV/Auth/Backend/IMAP.php
c0d3z3r0: It wasn't

lib/DAV/Auth/Backend/IMAP.php
c0d3z3r0: ... added a newline to match any other methods. I can move this to an own commit if you want me to.

staabm:
Perfect, looks fine tomme then.

lib/DAV/Auth/Backend/IMAP.php
```diff
@@ -37,10 +30,9 @@ class IMAP extends AbstractBasic {
      * @param string $imap_host
```
staabm: Docs should reflect your change

c0d3z3r0: Yep, done

staabm:
@sebastian-schmid could you verify this fix works for you?

@bakerjalexander would you also mind giving this changes a try?

sebastian-schmid:
yes, works great

Remember that $imap_host and $imap_flags are now one variable $mailbox and the whole string must be given as one now. Not seperated to two strings as before. I think this is the more obvious and PHP IMAP like way.

As this IMAP Auth backend was not released yet this change should not cause problems.

staabm:
Thank you guys

evert:
Great work

PR #916, opened 2016-11-28
Allow asset delivery from within phar files.

tobiastom:
I created a minimal server that provides the Cal|CardDAV functionality for me. I want to deploy it though a single phar file. For me that has the advantage that composer is not needed to be installed and that I can easily update it (to a tested version).

I can see that the check might not be optimal and I'm very open to suggestions to make it better.

### DIFF
--- a/lib/DAV/Browser/Plugin.php
+++ b/lib/DAV/Browser/Plugin.php
@@ -561,10 +561,20 @@ HTML;
         if (strpos($path, '/../') !== false || strrchr($path, '/') === '/..') {
             throw new DAV\Exception\NotFound('Path does not exist, or escaping from the base path was detected');
         }
-        $realPath = realpath($path);
-        if ($realPath && strpos($realPath, realpath($assetDir)) === 0 && file_exists($path)) {
-            return $path;
+
+        $scriptFilename = realpath($_SERVER['SCRIPT_FILENAME']);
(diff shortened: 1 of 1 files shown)


staabm:
I remember that phar based apps are very slow...  did you check that?

Composer itself is not a deployment tool, so it shouldnt be required at all to deploy your source into a production envirnment.

tobiastom:
The answer is of course: it depends. The speed inside the phar file is by far good enough for me. 

When used with an accelerator, I didn't see any difference to plain PHP code.

For me it's just making it work with your code base. Not about recommend it as a go to solution for everyone. ;)

staabm:
I have no opinion on the change itself, but at least we need a test to cover it ;-)

evert:
I don't really want to accept this patch, mainly because I never use `.phar`, nor know anyone that does.  We don't officially support this. This leaves me with a feature I'm not sure if it works, but it is something I need to maintain version over version. No tests doesn't help this.

So since you probably have some custom script to create the .phar build. why don't you just subclass the browser plugin and override that function? That gives you a super clean upgrade path specific to your situation. We don't break that function in minor versions, and if we do in a major version, we really clearly document it. It's how this feature is intended, so please just use that!

tobiastom:
Thank you @evert for your open words. I actually don't know anyone using phar deployment except myself, so you are most likely right. 

I'll go with the sub-class approach.

PR #941, opened 2017-02-01
Don't throw a Notice in calculateUri when passing an empty uri

icewind1991:
When a dav server listening on the root of a domain receives a request to the root of that domain, the request $uri will be an empty string.

Besides the one Notice that was thrown the rest of the implementation works correctly in that case ($uri = '', $baseUri = '/')

### DIFF
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -565,7 +565,7 @@ class Server implements LoggerAwareInterface, EmitterInterface {
      */
     function calculateUri($uri) {
 
-        if ($uri[0] != '/' && strpos($uri, '://')) {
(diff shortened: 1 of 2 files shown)


lib/DAV/Server.php
```diff
@@ -565,7 +565,7 @@ function getRequestUri() {
      */
     function calculateUri($uri) {
 
-        if ($uri[0] != '/' && strpos($uri, '://')) {
+        if ($uri != '' && $uri[0] != '/' && strpos($uri, '://')) {
```
staabm: which uri do you expect `calculateUri` to return for this new case? please add a unit test for it.

icewind1991: Added test

PR #946, opened 2017-02-15
Fix parsing bug for some CardDAV reports.

### LINKED ISSUE #943: Contacts does not sync
My Contacts seems not to sync. However, i see the Adressbookfolder in my Clients, but no Contacts are in there. If i use a Browser, i see all the Contacts. From Nginx i get the following line:

 "REPORT /sdav/addressbookserver.php/addressbooks/username@server.com/e5198823-9293-4b10-9242-cb0d40740189/ HTTP/1.1" 500 389 "-" "MSFT-WIN-3/10.0.14393 (gzip)"

In this Case, Windows Client will get the Contacts, the same is going on for eM Client. Calendar works fine.

evert:
Fixes #943

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,16 @@
 ChangeLog
 =========
 
+3.2.2 (2017-02-14)
+------------------
+
+* #943: Fix CardDAV XML reporting bug, which was affecting several CardDAV
+  clients. Bug was introduced in 3.2.1.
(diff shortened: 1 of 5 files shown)


DeepDiver1975:
Oooops .... sorry - this way me :wink:

evert:
yup :P no problem. Should have read the PR better

DeepDiver1975:
:+1:

PR #951, opened 2017-03-14
Throw exception if FSExt\Directory::moveInto() fails

NyroTheCat:
The node is moved by php rename(). However, if this fails (e.g. due to missing filesystem permissions) an exception is thrown.


evert:
Under normal circumstances a failed rename because of say, permissions, should already trigger a PHP error, which should trigger a PHP exception. Therefore I don't think this is a needed change. If you can write a unittest for this and it works, I think it's a good idea, but I think you'll find that in the unittests, the line where you introduce the exception is in fact, unreachable.

PR #965, opened 2017-05-17
Add reader->contextUri initialisation to parse function

tbille:
**Actual behavior :**
The build is actually failing because of the `testDeserialize` of the `/tests/Sabre/CardDAV/Xml/Request/AddressBookMultiGetTest.php`.

This is because the contextUri of the reader is not initialized at his creation.

https://travis-ci.org/fruux/sabre-dav/jobs/221757014

**Expecting behavior**
The tests should pass. :)


staabm:
Thx

PR #966, opened 2017-05-22
As a sharee with write access I want to invite attendees

tbille:
**Actual behavior**
As a sharee that has write access to a calendar. If an event is created by the sharer I can't add attendees. The attendees aren't notified, and the event is not created in their calendar.

**Expected behavior** 
As a sharee with write acces to a calendar, I want to add attendees to an event created by the sharer.

**Notes**
This pull request requires this pull request on the vobject project to be merged: https://github.com/fruux/sabre-vobject/pull/378


DeepDiver1975:
I can hardly judge on this .... will close as it anyhow is conflicting and would need a fresh restart

PR #967, opened 2017-05-23
Update Client.php

Xpndable:
Noted that on older WebDAV servers, a self closed d:prop will return a 400 Bad Request when no properties are provided.  Added modification to check for properties, if none found instead use d:allprop.


DeepDiver1975:
test is missing ....

PR #974, opened 2017-06-25
Empty default timezone with freeBusyQueryReport

lbeziaud:
If the default timezone for the calendar is an empty string, `VObject\Reader` will return `EofException`. Why the timezone is an empty string should be investigated! This fixes https://github.com/fruux/Baikal/issues/697


DeepDiver1975:
Test missing .....

PR #981, opened 2017-07-12
Add check for Lock-Token in request header.

datatrippers:
Only check for Lock-Token is there are $mustLocks left.
This applies for PUT requests, which used to fail when saving from Word 2016


staabm:
could you add a unit test to show your issue?

datatrippers:
I've copied the MSWord unit test and updated the PUT request with the LockToken header.

Attached you'll find the put request

[PUT request Word2016.txt](https://github.com/fruux/sabre-dav/files/1142496/PUT.request.Word2016.txt)

PVince81:
@davitol this explains why it didn't work back then when testing and where you had to switch to using "If"

DeepDiver1975:
word202 :zzz: -> close

PR #1006, opened 2017-10-06
fix "count(): Parameter must be an array or an object that implements…

remicollet:
… Countable" (7.2)


DeepDiver1975:
Tested - related unit tests are passing ... but we should merge this into master branch

PR #1011, opened 2017-10-18
Fix an inconsistency with RFC 7230 3.3.3 which clearly states:

jkimalane:
A server MUST NOT send a Content-Length header field in any response
with a status code of 1xx (Informational) or 204 (No Content)

This violation breaks applications behind OpenBSD relayd (version 1.61 and up) load balancer running http protocol handler as the relayd’s http protocol hendler implementation very closely follows RFC. This issue appears with HTTP method DELETE but other method, which return 204 are affected as well.

### DIFF
--- a/lib/DAV/CorePlugin.php
+++ b/lib/DAV/CorePlugin.php
@@ -288,8 +288,7 @@ class CorePlugin extends ServerPlugin {
         $this->server->emit('afterUnbind', [$path]);
 
         $response->setStatus(204);
-        $response->setHeader('Content-Length', '0');
-
+        
         // Sending back false will interrupt the event chain and tell the server
         // we've handled this method.
         return false;
@@ -514,7 +513,6 @@ class CorePlugin extends ServerPlugin {
                 return false;
             }
 
-            $response->setHeader('Content-Length', '0');
             if ($etag) $response->setHeader('ETag', $etag);
             $response->setStatus(204);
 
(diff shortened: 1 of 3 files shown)


evert:
These headers were originally actually added to solve issues with some clients expecting a content-length, although I don't remember which had problems with this. (long time ago)

jkimalane:
For example nextCloud (versions 10, 11, 12) does not work correctly without this patch when application servers are behind OpenBSD relayd (http protocol handler)

evert:
I understand @jkimalane , but without knowing what _will_ break with this change, it makes me a little worried. So far you're the first person to report this in the 10 years of this project so my assumption would be that you're more likely to be a minority than the users that will experience issues if this patch does get accepted.

jkimalane:
in relayd relay_http.c code there is following section (http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.sbin/relayd/relay_http.c?rev=1.67&content-type=text/x-cvsweb-markup) :

if (desc->http_method == HTTP_METHOD_RESPONSE && (
			    ((desc->http_status >= 100 &&
			    desc->http_status < 200) ||
			    desc->http_status == 204))) {
				relay_abort_http(con, 500,
				    "Internal Server Error", 0);
				goto abort;
			}
This issue appears always when multiple files are selected in nextCloud web interface and those selected files are deleted. In WUI "connection lost" error message is displayed as the relayd returns rc 500. Although DELETE transaction itself is successful and selected files are removed.

Regarding rareness of this issue, I assume that architecture we are using is not so common. For example this issue, as far as I remember, does not occur with F5. But as I said before, BSD's implementation follows RFC to the word as it clearly states that 204 must not return "Content-Length"

cb8:
The fix might was needed because of nginx: ab56aeac76

evert:
Thanks for the research @cb8 ! Sounds like this ticket should be closed.

PR #1020, opened 2017-12-20
Some notes on the release process.

evert:
This is mostly off memory, and I wrote it a bit quick. Let me know if
you have questions, happy to answer them.


DeepDiver1975:
Let's merge this - we will see how far we get once we are brave enough to do a release ;-)

PR #1023, opened 2018-01-05
Update CardDAV Plugin.php

tester19823:
Use of strict_types causes variable type error in line 729 of CardDAV Plugin.php if the value is NULL. Appropriate casting of variable into string is needed, i.e.:
`list($part) = explode(';', (string)$contentType);`


DeepDiver1975:
something went totally wrong here -> close

Issue #1026, opened 2018-01-19
Event is not created in attendee's calendar when iTIP messages are delivered delivered locally


stellingsimon:
We noticed that *sometimes*, in a SabreDAV system using scheduling and sharing features, certain recipients never receive their invitations.

We tracked this down to the following scenario:

Assume the following setting:

 * an organizer O: `principals/o`
 * a resource R: `principals/r`
 * an attendee A: `principals/a`

Further assume that the attendee A has `READ_WRITE` access to the resource's calendar sheet through a shared calendar instance.

The organizer creates an event with `R` and `A` listed as `ATTENDEE`s.

The iTIP broker will distribute the invitation for `R` first and place it in its calendar. It then distributes the invitation for `A`, but since `A` can read `R`'s calendar, it will find the event there and not create a second object in `A`'s calendar. This is due to a faulty implementation of `getCalendarObjectByUID`:

        /**
         * ...
         * This method should only consider * objects that the principal owns, so
         * any calendars owned by other principals that also appear in this
         * collection should be ignored.
         * ...
         */
        function getCalendarObjectByUID($principalUri, $uid) {

            $query = <<<SQL
    SELECT
        calendar_instances.uri AS calendaruri, calendarobjects.uri as objecturi
    FROM
        $this->calendarObjectTableName AS calendarobjects
    LEFT JOIN
        $this->calendarInstancesTableName AS calendar_instances
        ON calendarobjects.calendarid = calendar_instances.calendarid
    WHERE
        calendar_instances.principaluri = ?
        AND
        calendarobjects.uid = ?
    SQL;

            $stmt = $this->pdo->prepare($query);
            $stmt->execute([$principalUri, $uid]);

            if ($row = $stmt->fetch(\PDO::FETCH_ASSOC)) {
                return $row['calendaruri'] . '/' . $row['objecturi'];
            }

        }

As the comment already indicates, this method is missing a condition `calendar_instances.access = 1` (1 is the code for `OWNER`).

DeepDiver1975:
@stellingsimon mind submitting this as a pull request? THX

stratmaster:
I've noticed the same issue. So, would it be possible to merge this pull request #1032?

PR #1032, opened 2018-01-22
Fix for #1026: return only calendar objects owned by principal itself

stellingsimon:
Align the implementation of `getCalendarObjectByUID` with its documentation. 

I'd love to provide a unit test for #1026 that proves this actually fixes it, but I have zero experience with PHP development, so this would take prohibitively long. I did test it manually though (in 3.2.2)!

Thank you!


evert:
I think this looks pretty reasonable!

staabm:
Sorry for waiting so long here. The PR was duplicated in the meantime https://github.com/sabre-io/dav/pull/1181

PR #1036, opened 2018-02-08
Add codecov.io integration

DeepDiver1975:
## ToDo:
- [x] only enable coverage generation selectively - the execution time is extremly higher with coverage generation
- [ ] maybe use phpdbg for coverage runs - see https://hackernoon.com/generating-code-coverage-with-phpunite-and-phpdbg-4d20347ffb45
- [x] make decision to enable this at all

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ env:
     - SABRE_MYSQLPASS=""
     - SABRE_MYSQLDSN="mysql:host=127.0.0.1;dbname=sabredav_test"
   matrix:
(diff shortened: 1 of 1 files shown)


evert:
sweet!

staabm:
like it!

> maybe use phpdbg for coverage runs

As far as I remember there are some issues with php 7.2 and this feature right now, therefore I would stick with xdebug for now.

> only enable coverage generation selectively - the execution time is extremly higher with coverage generation

👍

Hywan:
:+1:

PR #1045, opened 2018-03-06
Support for custom copy operations

cb8:
This patch adds support for custom copy operations (in the same spirit as `IMoveTarget`). It helps especially in cases when users copy large file trees or when the underlying system supports some kind of copy-on-write technique.

### DIFF
--- a/lib/DAV/FSExt/Directory.php
+++ b/lib/DAV/FSExt/Directory.php
@@ -12,7 +12,7 @@ use Sabre\DAV\FS\Node;
  * @author Evert Pot (http://evertpot.com/)
  * @license http://sabre.io/license/ Modified BSD License
  */
-class Directory extends Node implements DAV\ICollection, DAV\IQuota, DAV\IMoveTarget {
+class Directory extends Node implements DAV\ICollection, DAV\IQuota, DAV\IMoveTarget, DAV\ICopyTarget {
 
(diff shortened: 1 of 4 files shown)


lib/DAV/FSExt/Directory.php
```diff
@@ -208,4 +208,33 @@ function moveInto($targetName, $sourcePath, DAV\INode $sourceNode) {
...
+     * @return bool
+     */
+    function copyInto($targetName, $sourcePath, DAV\INode $sourceNode) {
+
+        // We only support FSExt\File objects for now.
+        if (!$sourceNode instanceof File) {
+            return false;
+        }
+
+        copy($sourceNode->path, $this->path . '/' . $targetName);
```
staabm: I dont get the idea behind this implmentation here. why does it save something?
(having custom copy operations makes sense to me though)

could you elaborate a bit?

cb8: After further investigation, this implementation does not achieve much indeed. I assumed PHP would have some kind of shortcut to copy files. However, it seems that it just [creates two streams](https://github.com/php/php-src/blob/b7940d368fa627c1ae10f67f03cddd3ce764de5e/ext/standard/file.c#L1774-L1784) and copies them. A patch is available in #1046.

PR #1046, opened 2018-03-07
Remove apparent optimization of copy operations

cb8:
In `FSExt\Directory`, the copy() function will do the same as we would in Tree::copyNode(). Therefore, we can remove this method to avoid any confusion why it was implemented. Thanks to @staabm for noticing this. I assumed PHP would apply some tricks instead of just creating two streams.


staabm:
Thx

Issue #1050, opened 2018-03-12
tag a new releaese


staabm:
we should tag a new release with up2date dependencies for all the sabre/* dependencies

staabm:
@DeepDiver1975 maybe I ask you to get a release out for sabre/dav ?

DeepDiver1975:
release on 3.3 or 3.2 or both? ;-)

we shall discuss what from master is worth merging into either of these branches

staabm:
I feel the most usefull way would be to create a new feature release branch (so 3.3), merging all non-breaking master-only changes into it, update the dependencies and release 3.3.

If time allows we could also create another 3.2.* just with updated vendors.

DeepDiver1975:
we already have a branch 3.3 - it's diff to master looks like this: https://github.com/sabre-io/dav/compare/3.3...master

DeepDiver1975:
I'll open a PR tonight to merge back changes from master to 3.3 branchs - include updating changelog and so one

do we want to wait for the changes as proposed by @icewind1991 in his to latest PRs ?

staabm:
> we already have a branch 3.3 - it's diff to master looks like this: 3.3...master


ohh you are right. didn't realized we have a 3.3 branch already, since no 3.3.* release is available yet.



> I'll open a PR tonight to merge back changes from master to 3.3 branchs - include updating changelog and so one

that would be awesome! thx.

> do we want to wait for the changes as proposed by @icewind1991 in his to latest PRs ?

only in case you, or @evert is able to review and approve it in time. in case this is not possible, we should release without the PR.

gharlan:
https://github.com/sabre-io/dav/blob/master/composer.json#L21
sabre/http 5.0 should be released, too.

DeepDiver1975:
will do a sabre release session tonight ....

evert:
Yea please kill the 3.2 branch =) 3.3 was ready a long time ago imho and it's pretty much compatible with 3.2.

staabm:
@DeepDiver1975 do you have time to create a new alpha6.. ? maybe we could also declare it stable since we had a lot of alphas already (or maybe do a beta?)

DeepDiver1975:
Sure .... will submit a PR asap. I'm find building a beta.

PR #1058, opened 2018-05-23
Don't open the file on HEAD request

icewind1991:
Opening the file, even if the stream doesn't get read, is likely to lead to unneeded overhead if the body get's throw away afterwards anyway.

### DIFF
--- a/lib/DAV/CorePlugin.php
+++ b/lib/DAV/CorePlugin.php
@@ -82,14 +82,18 @@ class CorePlugin extends ServerPlugin {
 
         if (!$node instanceof IFile) return;
 
-        $body = $node->get();
+        if ($request->getHeader('X-Sabre-Original-Method') === 'HEAD') {
+            $body = '';
+        } else {
+            $body = $node->get();
 
-        // Converting string into stream, if needed.
-        if (is_string($body)) {
-            $stream = fopen('php://temp', 'r+');
-            fwrite($stream, $body);
-            rewind($stream);
-            $body = $stream;
+            // Converting string into stream, if needed.
+            if (is_string($body)) {
+                $stream = fopen('php://temp', 'r+');
+                fwrite($stream, $body);
+                rewind($stream);
+                $body = $stream;
+            }
         }
 
         /*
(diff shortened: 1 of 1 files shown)


staabm:
Did you run any benchmark on this change?

icewind1991:
I'm not sure how much meaningful benchmarks you can get since it will heavily depend on the filesystem implementation.

In some Nextcloud setups, opening the file will require sending http requests to external servers which would obviously cause a large overhead, while in others it will just lead to opening a local file which should be minimal in overhead.

lib/DAV/CorePlugin.php
```diff
@@ -248,6 +252,8 @@ function httpHead(RequestInterface $request, ResponseInterface $response) {
         // and dropping the response body.
         $subRequest = clone $request;
         $subRequest->setMethod('GET');
+        // signal to the GET request handler that it doesn't need to read the file
+        $subRequest->setHeader('original-method', 'HEAD');
```
staabm: Name-wise I would prefer something like ˋX-Sabre-Original-Methodˋ

staabm:
Except the naming of the new header this LGTM

lib/DAV/CorePlugin.php
icewind1991: done

lib/DAV/CorePlugin.php
```diff
@@ -248,6 +252,8 @@ function httpHead(RequestInterface $request, ResponseInterface $response) {
         // and dropping the response body.
         $subRequest = clone $request;
         $subRequest->setMethod('GET');
+        // signal to the GET request handler that it doesn't need to read the file
+        $subRequest->setHeader('X-Sabre-Original-Method', 'HEAD');
 
         try {
             $this->server->invokeMethod($subRequest, $response, false);
```
staabm: Set the body to an empty string should no longer be necessary here?

lib/DAV/CorePlugin.php
```diff
@@ -248,6 +252,8 @@ function httpHead(RequestInterface $request, ResponseInterface $response) {
         // and dropping the response body.
```
staabm: This comment is no longer correct, at least the last part

icewind1991:
Updated the comments and removed the unneeded "set empty body"

staabm:
Thx

DeepDiver1975:
- [ ] we need to add this to the changelog
- [ ] make decision about merging this to 3.3 and /or 3.2

staabm:
I would agree merging this feature into a 3.3 release (with others only present in master which are non-breaking)

DeepDiver1975:
> I would agree merging this feature into a 3.3 release (with others only present in master which are non-breaking)

lets discuss this in here - https://github.com/sabre-io/dav/issues/1050#issuecomment-394346496

DeepDiver1975:
@icewind1991 there is one scenario where this optimization fails to behave properly:

- a node is existing and the call to $node = $this->server->tree->getNodeForPath($path); is returning the node
- but when calling get an excpetion is thrown because the file physically does not exist or the user has no access or the resource is physically locked or ....

in the scenario above after the optimization the HEAD request will return 200. before the proper status

PR #1059, opened 2018-05-24
Allow nesting of trees

icewind1991:
Makes `Tree` implement `ICollection` and modify `getNodeForPath` to pass the request to nested trees.

My main use case for this in Nextcloud is that for part of the tree we can optimize `getNodeForPath` by directly getting the target node instead of going trough all parent nodes, but there is currently no clean way to apply this optimization to only a part of the dav tree.
With nested trees we could put this optimization in a separate `Tree` subclass and nest it into the "main" tree.

### DIFF
--- a/lib/DAV/Tree.php
+++ b/lib/DAV/Tree.php
@@ -14,7 +14,7 @@ use Sabre\Uri;
  * @author Evert Pot (http://evertpot.com/)
  * @license http://sabre.io/license/ Modified BSD License
  */
-class Tree {
+class Tree implements ICollection {
 
     /**
      * The root node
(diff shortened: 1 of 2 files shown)


evert:
I've been considering doing this exact thing for a while, because it kind of allows for a sort of 'mounting' type of system.

However,  the methods having multiple purposes (`delete()` vs `delete($path)`) makes me feel a bit uncomfortable.

So one thing I'm wondering is, could it make sense to instead just add a new interface that only defines the `getNodeForPath()` function. (or maybe `getChildByPath()`, idk), or does that not satisfy your use-case?

icewind1991:
I'm also not a fan of the method name conflicts.

A separate interface for just the `getNodeForPath` would probably cover the specific use case I started the work for but having full nesting of trees would most likely be useful in future cases (I'm fairly sure it would allow us to do some good code reuse)

icewind1991:
True.

I would go for `getNodeForPath` instead of `getChildByPath` or something similar because then `Tree` can simply implement the new interface.

I'll look into implementing this tomorrow

DeepDiver1975:
@icewind1991 let's revive this once work continues on #1060

DeepDiver1975:
@icewind1991 rebase and move on? :wink:

DeepDiver1975:
feel free to reopen if this is still of interest

PR #1060, opened 2018-05-25
Add INodeByPath interface to allow optimizing getNodeForPath

icewind1991:
As discussed in https://github.com/sabre-io/dav/pull/1059, doing this in a separate PR so the other can stay open for the "nested tree" discussion.

A few points of detail

- `INodeByPath` does not extend `ICollection` so `Tree` can implement the interface without further changes, having `Tree` implement this does not really provide any benefit atm though and was mostly done to support future tree nesting. It might be better to change it to extend `ICollection` and make `Tree` implement it if we ever make a collection.
- `INodeByPath::getNodeForPath` can return `null`, I fear that requiring any collection implementing the interface to be able to retrieve **all** nodes within it's path would prevent the ability to cleanly allow nesting collections that implement the interface

### DIFF
--- a/lib/DAV/INodeByPath.php
+++ b/lib/DAV/INodeByPath.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabre\DAV;
+
+/**
+ * INodeByPath.
+ *
+ * This interface adds a tiny bit of functionality to collections.
+ *
+ * Getting a node that is deep in the tree normally requires going trough each parent node
+ * which can cause a significant performance overhead.
+ *
+ * Implementing this interface allows solving this overhead by directly jumping to the target node.
+ *
+ * @copyright Copyright (C) Robin Appelman (https://icewind.nl/)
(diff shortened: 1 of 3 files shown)


staabm:
I like the general idea. I dont know enough about this code parts, so dont expect I will merge it. 

Hopefully @evert or somebody else will have enough time to get this merged/reviewed

skjnldsv:
Bump?

evert:
I'm 4 years late but I do think this is a good idea :S

DeepDiver1975:
@icewind1991 it has been a while ... :see_no_evil: .... wana revive this with a rebase? THX

DeepDiver1975:
THX @icewind1991

DeepDiver1975:
Test are failing due to a change in the pgsql schema .....

DeepDiver1975:
> Test are failing due to a change in the pgsql schema .....

fixed by https://github.com/sabre-io/dav/pull/1509

DeepDiver1975 (approved):

PR #1074, opened 2018-08-26
Add event to allow inspecting and changing multipart responses

icewind1991:
Use case for this is allowing caching/pagination of multipart responses.

### DIFF
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -1635,6 +1635,8 @@ class Server implements LoggerAwareInterface, EmitterInterface
      */
     public function generateMultiStatus($fileProperties, $strip404s = false)
     {
+        $this->emit('beforeMultiStatus', [&$fileProperties]);
+
         $w = $this->xml->getWriter();
         if (self::$streamMultiStatus) {
             return function () use ($fileProperties, $strip404s, $w) {
--- a/tests/Sabre/DAV/ServerEventsTest.php
+++ b/tests/Sabre/DAV/ServerEventsTest.php
@@ -12,6 +12,8 @@ class ServerEventsTest extends AbstractServerTestCase
 
     private $exception;
 
+    private $fileProperties;
+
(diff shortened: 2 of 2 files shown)


DeepDiver1975:
@icewind1991 Can I ask you for an example code snippet on how this event can be used?

I'm missing some logic which tells the code on what to do after the event was take care of or not.

THX

icewind1991:
My current use case for this is here: https://github.com/nextcloud/server/pull/10880/files#diff-a8ef54ad5b45404480feafc1de1eea28R62 where it stores a copy of the mutlipart listing result and limits the output returned.

DeepDiver1975:
> My current use case for this is here: https://github.com/nextcloud/server/pull/10880/files#diff-a8ef54ad5b45404480feafc1de1eea28R62 where it stores a copy of the mutlipart listing result and limits the output returned.

I see you are using it for caching the dataset for later reuse - got it.

Why are you passing the properties by ref then?

DeepDiver1975:
Got it - THX

I was having trouble in the past as well with the missing capabilities to change the behavior of generateMultiStatus

What about a more generic approach where the whole method can be overwritten?

like emitting a generateMultiStatus event and in case not handled we fall back to the default implementation.

My use case was to chnage the content type from xml to json ...

@staabm what's our point on this? THX

icewind1991:
The ability to overwrite the method would also work yes, this was just the minimal approach I found

DeepDiver1975:
@icewind1991 how do we continue with this one?

icewind1991:
Rebased.

I would still like to have this merged as I believe that it provides a useful option for plugin authors at minimal cost.

staabm (approved):

phil-davis (approved):
LGTM

I will have a look through other pending `sabre` repo PRs over the next day or so, and probably make some releases "soon".

phil-davis:
Ported to the major version 4 release series in PR #1549

PR #1091, opened 2018-10-05
Add phpstan to travis

DeepDiver1975:


lib/CalDAV/Plugin.php
```diff
@@ -3,6 +3,7 @@
 namespace Sabre\CalDAV;
 
 use DateTimeZone;
+use Sabre\CalDAV\Xml\Request\CalendarMultiGetReport;
```
staabm: Hmm is it really missing?

lib/CalDAV/SharingPlugin.php
```diff
@@ -136,8 +136,7 @@ function propFindEarly(DAV\PropFind $propFind, DAV\INode $node) {
                 }
 
                 return new Xml\Property\Invite(
-                    $node->getInvites(),
-                    $ownerInfo
+                    $node->getInvites()
```
staabm: When ownerInfo really is not required here, we should also delete some of the logic above this line

phpstan.neon
```diff
@@ -0,0 +1,6 @@
+parameters:
+  level: 0
+  bootstrap: %currentWorkingDirectory%/vendor/autoload.php
+
+  excludes_analyse:
+    - %currentWorkingDirectory%/lib/DAV/Auth/Backend/IMAP.php
```
staabm: Why excluded?

lib/CalDAV/Plugin.php
DeepDiver1975: It is used in PHPDoc as type annotation

lib/CalDAV/SharingPlugin.php
DeepDiver1975: Invite:__construct only take one arg .... I will review the code once more .... THX for the pointer

phpstan.neon
DeepDiver1975: Will add comment - because the imap module is only a suggested dependency .... need to have a second look ...

lib/DAVACL/Xml/Request/ExpandPropertyReport.php
```diff
@@ -69,7 +69,7 @@ static function xmlDeserialize(Reader $reader) {
      * {DAV:}property elements.
      *
      * @param array $elems
-     * @return void
+     * @return array
```
staabm: Hmm `string[]` ?

DeepDiver1975: the returned array could also hold an array of strings itself .... just given by the code logic

DeepDiver1975:
@staabm final review please - all should be sorted out now. THX

lib/CalDAV/Schedule/SchedulingObject.php
```diff
@@ -46,6 +46,7 @@ class SchedulingObject extends \Sabre\CalDAV\CalendarObject implements IScheduli
      * @param array $objectData
      */
     function __construct(Backend\SchedulingSupport $caldavBackend, array $objectData) {
+        parent::__construct($caldavBackend, [], $objectData);
 
         $this->caldavBackend = $caldavBackend;
```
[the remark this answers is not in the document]
DeepDiver1975: yes - I need to have a closer look. The PHPDoc type annotation is different.

DeepDiver1975:
@staabm ready for some final review? THX

staabm:
Hmm seems build is failling

DeepDiver1975:
@staabm fixed now

staabm:
Cool thx

PR #1092, opened 2018-10-08
Remove sabre/cs

DeepDiver1975:


DeepDiver1975:
@staabm any idea why the prettyci check is not popping up? :confused:

staabm:
@DeepDiver1975 thx check seems to run now

DeepDiver1975:
> @DeepDiver1975 the check seems to run now

We need to review the print out and see if we agree or need to adapt the config

staabm:
as stated elsewhere: I dont prefer any style and would welcome any "standard open source style", like e.g. the symfony you configured.

evert also wasn't fixed on the old style.

in case this fits for you we can go ahead.

PR #1093, opened 2018-10-09
cought with litmus props test propfind_invalid2

DeepDiver1975:


staabm:
lets wait for a release of sabre/xml which contains sabre-io/xml#149 so we can also get the fix for the problem into this PR.

composer.json
```diff
@@ -17,7 +17,7 @@
         "php": ">=7.0.0",
         "sabre/vobject": "^4.2.0-alpha1",
         "sabre/event" : "^5.0",
-        "sabre/xml"  : "^2.0.0",
+        "sabre/xml"  : "^2.1.1",
```
staabm: we shouldn't rais this minimum version. composer will automatically pick the latest version which is compatible.

the version mentioned in the composer.json of our libs should be as low as possible, to support the widest range possible.

DeepDiver1975: but if we know that 2.1.0 is broken - we better force this dependency. Otherwise users will run into the same issue and finding out about the reason is not trivial.

I prefer to be better safe then sorry ...

staabm: every release of a software product will fix "some things" but only few of them will affect most users.
its not our job to tell people to use the latest versions of the used libraries.
in case we are too strict, people might even get a "composer is not able to find a compatible set of depdencies" error, so they can't use the software at all.

I feel this as a good general rule.

in our concrete case here the fix in 2.1.1 is feels like a edge-case and most users will not produce xml like that.

do you think error in the namespace are a common problem?

DeepDiver1975:
@staabm I reverted the change to composer.

tests/Sabre/DAV/Xml/ServiceTest.php
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Sabre\DAV\Xml;
+
+use PHPUnit\Framework\TestCase;
+
+class ServiceTest extends TestCase
+{
+
+    /**
+     * @expectedException \Sabre\Xml\LibXMLException
```
staabm: we need to skip this test when `LOWEST_DEPS` ENV var is empty

DeepDiver1975: Hint on how to do this is welcome ..... but I have to say that this is fishy from my pov ....

Please note that the issue we are trying to fix is a regression - it was working properly in previous dav versions ......

staabm: hmm ok, this is new information. shouldn't we backport the fix into 2.0.x then?

DeepDiver1975: let me see what I can do ......

staabm: cool. thank you very much.

merged and restarted the travis build.

staabm:
thx

DeepDiver1975:
We did not yet sort out travis failure for lowest_deps ....

staabm:
Will be fixed in 
https://github.com/sabre-io/dav/pull/1095

PR #1094, opened 2018-10-09
Apply php-cs-fixer rule set @symfony

DeepDiver1975:


staabm:
I guess this needs a rebase. then we are are getting green again :-)

staabm:
Thx

PR #1098, opened 2018-10-12
Bugfix/caldav plugin strpos null

DeepDiver1975:


staabm:
Does it require a 3.x backport?

DeepDiver1975:
> Does it require a 3.x backport?

I don't think it's required because 3.x does not use strict types. Never the less backporting will not hurt

DeepDiver1975:
Looks like I need to adjust some tests .....

staabm:
> Looks like I need to adjust some tests .....

hmm travis passes, no?

DeepDiver1975:
> hmm travis passes, no?

travis is fine - but the coderage dropped ....

DeepDiver1975:
@staabm coverage is fine now. Merge? THX

lib/DAV/Sharing/Plugin.php
```diff
@@ -172,7 +172,7 @@ public function httpPost(RequestInterface $request, ResponseInterface $response)
         $contentType = $request->getHeader('Content-Type');
 
         // We're only interested in the davsharing content type.
-        if (false === strpos($contentType, 'application/davsharing+xml')) {
+        if (false === strpos((string) $contentType, 'application/davsharing+xml')) {
```
staabm: why do we need the string cast here, but not in the other cases?

staabm: Since all other places touched with this PR use a explicit null?check I would prefer having one here, too

DeepDiver1975:
@staabm ready now - THX

staabm:
Thx

PR #1108, opened 2018-10-23
Fix wrong return Type, backport #1106

micbar:
The PHP doc needs to be correct, phan tests fail.


staabm:
I dont think changes is worth beeing backported.

DeepDiver1975:
> I dont think changes is worth beeing backported.

we have been falling over this in the contact of owncloud and using phpstan ...

DeepDiver1975:
@micbar pretty ci is to be ignored on this branch - styles only apply to master branch as of now

micbar:
ok, thanks for the hint

staabm:
> we have been falling over this in the contact of owncloud and using phpstan ...

I see.

here we go.

PR #1125, opened 2018-12-10
Support lock timeout value Infinity

### LINKED ISSUE #1123: Lock timeout value is always "Second-"
See https://github.com/owncloud/core/issues/33778.

When debugging I noticed that even if the "Timeout" header is not specified, the XML response will still render "Second-" here https://github.com/sabre-io/dav/blob/master/lib/DAV/Xml/Property/LockDiscovery.php#L92.

If no timeout is specified the response should either default to the value "Infinite" or have some sensible default. The [RFC](https://tools.ietf.org/html/rfc4918#page-21) says that "The lifetime is suggested by the client when creating or refreshing the lock, but the server ultimately chooses the timeout value.  Timeout is measured in seconds remaining until lock expiration.".

@DeepDiver1975

DeepDiver1975:
fixes #1123

### DIFF
--- a/lib/DAV/Xml/Property/LockDiscovery.php
+++ b/lib/DAV/Xml/Property/LockDiscovery.php
@@ -89,7 +89,7 @@ class LockDiscovery implements XmlSerializable
                 $writer->endElement(); // {DAV:}lockroot
             }
             $writer->writeElement('{DAV:}depth', (DAV\Server::DEPTH_INFINITY == $lock->depth ? 'infinity' : $lock->depth));
-            $writer->writeElement('{DAV:}timeout', 'Second-'.$lock->timeout);
+            $writer->writeElement('{DAV:}timeout', (LockInfo::TIMEOUT_INFINITE === $lock->timeout ? 'Infinite' : 'Second-'.$lock->timeout));
 
             $writer->startElement('{DAV:}locktoken');
             $writer->writeElement('{DAV:}href', 'opaquelocktoken:'.$lock->token);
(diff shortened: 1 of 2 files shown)


lib/DAV/Xml/Property/LockDiscovery.php
```diff
@@ -89,7 +89,7 @@ public function xmlSerialize(Writer $writer)
                 $writer->endElement(); // {DAV:}lockroot
             }
             $writer->writeElement('{DAV:}depth', (DAV\Server::DEPTH_INFINITY == $lock->depth ? 'infinity' : $lock->depth));
-            $writer->writeElement('{DAV:}timeout', 'Second-'.$lock->timeout);
+            $writer->writeElement('{DAV:}timeout', (LockInfo::TIMEOUT_INFINITE === $lock->timeout ? 'Infinite' : 'Second-'.$lock->timeout));
```
staabm:
```suggestion
            $writer->writeElement('{DAV:}timeout', (LockInfo::TIMEOUT_INFINITE === $lock->timeout ? 'infinite' : 'Second-'.$lock->timeout));
```

Like for depth above?

DeepDiver1975:
Timeout RFC: https://tools.ietf.org/html/rfc4918#section-10.7

Depth RFC: https://tools.ietf.org/html/rfc4918#section-10.2

No idea why casing is different and why it is once called infinity and once Infinite ... :man_shrugging: :woman_shrugging:

evert:
That is confusing

staabm:
Thx

Issue #718, opened 2015-09-14
browse: show displayname instead of uri


mbiebl:
Version: 2.0 and later

I have the Browser Plugin enabled.
When I browse to http://<host>/addressbooks/<principal>/ it shows me the list of available addressbooks, but it uses the uri, previous releases showed the displayname.

In my case, I use UUIDs for the addressbooks, which aren't particularly readable.
It would be great, if one could configure whether to show the uri or the displayname, or maybe show both.

mbiebl:
It currently looks like this:
```
bc761905-7278-4bb2-a610-555952b7471b	Collection, Address book	
6c49a1ac-b0df-4656-975d-15269d43b902    Collection, Address book
```

mbiebl:
And browsing to a particular addressbook, the displayname is also not shown in "Properties" box.

evert:
This indeed got changed, because I was worried it was confusing in some instances. Have to think a bit about this...

mbiebl:
Fixed by 1760e86084cadc8731880fda89da6db69a9ab88a
Thus closing

PR #1136, opened 2019-02-08
Show displayName in browser

mbiebl:
See #718

### DIFF
--- a/lib/DAV/Browser/Plugin.php
+++ b/lib/DAV/Browser/Plugin.php
@@ -314,6 +314,10 @@ class Plugin extends DAV\ServerPlugin {
                     $lastMod = $subProps['{DAV:}getlastmodified']->getTime();
                     $html .= $this->escapeHTML($lastMod->format('F j, Y, g:i a'));
                 }
+                $html .= '</td><td>';
+                if (isset($subProps['{DAV:}displayname'])) {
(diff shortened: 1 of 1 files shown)


lib/DAV/Browser/Plugin.php
```diff
@@ -316,6 +316,10 @@ public function generateDirectoryIndex($path)
                     $lastMod = $subProps['{DAV:}getlastmodified']->getTime();
                     $html .= $this->escapeHTML($lastMod->format('F j, Y, g:i a'));
                 }
+                $html .= '</td><td>';
```
Hywan: I would have place the `</td><td>` inside the `if` condition, so that no empty cell is created.

Or maybe it is expected to get an empty cell?

mbiebl: @Hywan I thought about this as well. But if you look, you'll see that the other fields also unconditionally set `</td><td>`. I suspect the reason behind this is, that you'll always get the same number of columns per row, no matter if certain properties are set or not.

mbiebl:
Thanks for merging this so quickly, @DeepDiver1975 !

mbiebl:
@DeepDiver1975 would it be possible to cherry-pick this for 3.2 branch and make a 3.2.4 release?
I also noticed that the CHANGELOG.md in master does not contain all changes from the 3.2 branch, btw.

PR #1139, opened 2019-03-20
Fixes #1135 replaced continue with break in switch statement.

### LINKED ISSUE #1135: Error PHP "continue" targeting switch is equivalent to "break"
This error pops up for two files in the nextcloud log:

> 3rdparty/sabre/vobject/lib/Component/VCard.php#514
> 3rdparty/sabre/dav/lib/CalDAV/ICSExportPlugin.php#321

replacing the continue statement by "continue 2" solves this, but maybe it shoule be a break?

matejsvajger:
This is a bug fix for 3.2 branch, that fixes the issue with the use of continue statement inside a switch block in PHP 7.3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 ChangeLog
 =========
 
+3.2.4 (2019-03-20)
+------------------
+
+* #1135: Fix use of continue in switch for PHP7.3 compabiltiy
+
 3.2.3 (2018-10-19)
 ------------------
 
--- a/lib/DAV/Version.php
+++ b/lib/DAV/Version.php
@@ -14,6 +14,6 @@ class Version {
     /**
      * Full version number
      */
-    const VERSION = '3.2.2';
+    const VERSION = '3.2.4';
 
 }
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 ChangeLog
 =========
 
+3.2.4 (2019-03-20)
```
DeepDiver1975: please don't add a release date - this is done separatly

lib/DAV/Version.php
```diff
@@ -14,6 +14,6 @@ class Version {
     /**
      * Full version number
      */
-    const VERSION = '3.2.2';
+    const VERSION = '3.2.4';
```
DeepDiver1975: done separately as well

staabm:
> For php 7.2/7.3 support we have the upcoming version 4.0.0
> 
> I'm not sure if we want to add php 7.2/7.3 support to the 3.2 branch.
> If we do - we need to adjust the travis config as well

we should define a upper bound for the supported php version in the 3.2.x branch
https://github.com/sabre-io/dav/blob/6252622870258f818f39cb4a4af2fba9a7e7d1df/composer.json#L17

in case 3.2.x as is works on 7.2/7.3 and/or 7.4 (or only with warnings - but works) we should add them to .travis.yml.
In case we need to invest too much man power to get this compatibility, we should define the 7.1 (or whatever) php upper bound version and work on 4.* for the php compat

DeepDiver1975:
#1159 shows that there is much more to be dine to properly support php 7.2 and php 7.3

We will not invest time into this direction. Sorry

PR #1143, opened 2019-04-06
Update Directory.php - Fix folder (file) move issue if rename fails (…

richardleger:
…if $sourceNode->path is NULL)

The moveInto function shall always return the result value of the rename function instead of always true. So it would return false if the rename function fails (as $sourceNode->path seems NULL by default) and not always true as currently...


evert:
Thank you

PR #1145, opened 2019-04-26
Add sender name in invite mail headers

benrubson:
Hi,

This tiny PR adds the sender name in Cal invite mail headers.

Thank you 👍

PR #1150, opened 2019-04-27
Fix sending mail

ByteHamster:
These changes are needed to get mail working on my server. Publishing in case someone else has the same problem. I think this does no harm when merging for everyone.

    if ($iTipMessage->recipientName && $iTipMessage->recipientName != $recipient) {

My server silently fails when using `mail()` with a receiver that looks like  `a@b.c <a@b.c>`. It works fine when using `d <a@b.c>`. Looks like it does not accept using an email address as name.

    $subject = 'Invitation: '.$summary;

My server displays an error when not prepending a string: `mail() expects parameter 2 to be string, object given`

### DIFF
--- a/lib/CalDAV/Schedule/IMipPlugin.php
+++ b/lib/CalDAV/Schedule/IMipPlugin.php
@@ -112,7 +112,7 @@ class IMipPlugin extends DAV\ServerPlugin
         if ($iTipMessage->senderName) {
             $sender = $iTipMessage->senderName.' <'.$sender.'>';
         }
-        if ($iTipMessage->recipientName) {
+        if ($iTipMessage->recipientName && $iTipMessage->recipientName != $recipient) {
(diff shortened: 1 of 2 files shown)


benrubson:
Related to #1147 ? :)

ByteHamster:
Woah, it's the same PR :D Looks like you ran into the same problem when implementing https://github.com/sabre-io/Baikal/pull/787

benrubson:
Exactly !
Yours is better, with the added _Invitation_ label.
I then close mine ?

benrubson:
You also have #1145 which is in some kind related to this.

ByteHamster:
@staabm I added a unit test for the case where a wrong sender address is added. I can not reproduce the other error anymore. Do you prefer to remove it?

ByteHamster:
@staabm what do you mean with "Please go ahead"? I do not have merge permissions for this repo.

staabm:
I see :)

PR #1155, opened 2019-05-14
Updated copyright year

ByteHamster:


mbiebl:
Please don't do that. The copyright year should not be determined when the PHP script has been executed, but when it last has been modified.

evert:
The real shocker here is that it's been 3 years since somebody noticed it O_o

PR #1159, opened 2019-06-24
Test 3.2 with php 7.2 and 7.3 - and add composer max version for php …

DeepDiver1975:
…to 7.4


staabm:
because of a lot of travis errors, I guess it would be easiert to define the upper bound to be 7.1.

@DeepDiver1975 In case you are willing to work thru all the errors/problems I am fine with php 7.2/7.3 support for 3.2. I dont have time to do this myself

staabm:
one problem you will get will be to get the test-suite working across all supported php versions, because you need different phpunit major versions across the board:

https://phpunit.de/supported-versions.html

DeepDiver1975:
> one problem you will get will be to get the test-suite working across all supported php versions, because you need different phpunit major versions across the board:

yeah - it is not that easy ... I think we just go for a 7.1 max in composer for the 3.2 branch and we are done.

evert:
Why not just drop 7.1 as well? People will always have the option to use a slightly older version of this lib. This worked pretty well in the past.

staabm:
this PR is about the question whether to add php 7.2/7.3 into the latest stable release.

I guess what you are talking about is dropping php 7.1 from master branch?

evert:
Oops yes sorry.

staabm:
sabre/dav 3.2.x already supports 7.1

PR #1165, opened 2019-07-02
Fix Strict Error isTempFile

### LINKED ISSUE #1164: Strict Error TemporaryFileFilterPlugin
preg_match() expects parameter 2 to be string, null given

WorksDev:
- fix #1164

### DIFF
--- a/lib/DAV/TemporaryFileFilterPlugin.php
+++ b/lib/DAV/TemporaryFileFilterPlugin.php
@@ -172,6 +172,10 @@ class TemporaryFileFilterPlugin extends ServerPlugin
         // We're only interested in the basename.
         list(, $tempPath) = Uri\split($path);
 
+        if (null === $tempPath) {
+            return false;
+        }
+
         foreach ($this->temporaryFilePatterns as $tempFile) {
             if (preg_match($tempFile, $tempPath)) {
                 return $this->getDataDir().'/sabredav_'.md5($path).'.tempfile';
--- a/tests/Sabre/DAV/TemporaryFileFilterTest.php
+++ b/tests/Sabre/DAV/TemporaryFileFilterTest.php
@@ -99,6 +99,17 @@ class TemporaryFileFilterTest extends AbstractServer
         $this->assertEquals('Testing new file', stream_get_contents($this->response->body));
     }
 
+    public function testGetWithBrowserPlugin()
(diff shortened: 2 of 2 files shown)


lib/DAV/TemporaryFileFilterPlugin.php
```diff
@@ -173,7 +173,7 @@ protected function isTempFile($path)
         list(, $tempPath) = Uri\split($path);
 
         foreach ($this->temporaryFilePatterns as $tempFile) {
-            if (preg_match($tempFile, $tempPath)) {
+            if (preg_match($tempFile, $tempPath ?? '')) {
```
staabm: this case should be handled above the loop and we should early return then

staabm:
thx for working on a fix. please find my comments in the PR.

lib/DAV/TemporaryFileFilterPlugin.php
WorksDev: You're right about that. Changed. Thank you.

tests/Sabre/DAV/TemporaryFileFilterTest.php
```diff
@@ -4,6 +4,7 @@
 
 namespace Sabre\DAV;
 
+use Sabre\DAV\Browser\Plugin;
```
staabm:
```suggestion
use Sabre\DAV\Browser;
```

tests/Sabre/DAV/TemporaryFileFilterTest.php
```diff
@@ -99,6 +100,17 @@ public function testPutGet()
         $this->assertEquals('Testing new file', stream_get_contents($this->response->body));
     }
 
+    public function testGetWithBrowserPlugin()
+    {
+        $this->server->addPlugin(new Plugin());
```
staabm:
```suggestion
        $this->server->addPlugin(new Browser\Plugin());
```

staabm:
looks good, waiting for a green build. thx!

staabm:
awesome, thx

WorksDev:
@staabm is there already a planned date for a new release of the latest fixes? Thanks for the info in advance.

staabm:
lets see https://github.com/sabre-io/dav/issues/1177

PR #1169, opened 2019-07-06
Fixed null path error

ByteHamster:
When opening the http/get user interface on the `/` path, an error is displayed. All other pages work. Might be related to #1165. Where do all those `NULL` paths come from? Do you have an idea, @DeepDiver1975 @staabm? Instead of fixing in each plugin, we should probably fix this somewhere further up the call stack.

```
#0 baikal/vendor/sabre/dav/lib/DAV/Tree.php(58): trim(NULL, '/')
#1 baikal/vendor/sabre/dav/lib/CalDAV/Plugin.php(80): Sabre\DAV\Tree->getNodeForPath(NULL)
#2 baikal/vendor/sabre/dav/lib/DAV/Server.php(520): Sabre\CalDAV\Plugin->getHTTPMethods('')
#3 baikal/vendor/sabre/dav/lib/DAV/CorePlugin.php(825): Sabre\DAV\Server->getAllowedMethods('')
#4 baikal/vendor/sabre/dav/lib/DAV/Browser/PropFindAll.php(53): Sabre\DAV\CorePlugin->Sabre\DAV\{closure}()
#5 baikal/vendor/sabre/dav/lib/DAV/CorePlugin.php(827): Sabre\DAV\Browser\PropFindAll->handle('{DAV:}supported...', Object(Closure))
#6 baikal/vendor/sabre/event/lib/WildcardEmitterTrait.php(96): Sabre\DAV\CorePlugin->propFind(Object(Sabre\DAV\Browser\PropFindAll), Object(Sabre\DAV\SimpleCollection))
#7 baikal/vendor/sabre/dav/lib/DAV/Server.php(1053): Sabre\DAV\Server->emit('propFind', Array)
#8 baikal/vendor/sabre/dav/lib/DAV/Browser/Plugin.php(344): Sabre\DAV\Server->getPropertiesByNode(Object(Sabre\DAV\Browser\PropFindAll), Object(Sabre\DAV\SimpleCollection))
#9 baikal/vendor/sabre/dav/lib/DAV/Browser/Plugin.php(143): Sabre\DAV\Browser\Plugin->generateDirectoryIndex('')
#10 baikal/vendor/sabre/event/lib/WildcardEmitterTrait.php(96): Sabre\DAV\Browser\Plugin->httpGet(Object(Sabre\HTTP\Request), Object(Sabre\HTTP\Response))
#11 baikal/vendor/sabre/dav/lib/DAV/Server.php(464): Sabre\DAV\Server->emit('method:GET', Array)
#12 baikal/vendor/sabre/dav/lib/DAV/Server.php(241): Sabre\DAV\Server->invokeMethod(Object(Sabre\HTTP\Request), Object(Sabre\HTTP\Response))
#13 baikal/vendor/sabre/dav/lib/DAV/Server.php(309): Sabre\DAV\Server->start()
#14 baikal/Core/Frameworks/Baikal/Core/Server.php(123): Sabre\DAV\Server->exec()
#15 baikal/html/dav.php(61): Baikal\Core\Server->start()
#16 {main}
```

### DIFF
--- a/lib/CalDAV/Plugin.php
+++ b/lib/CalDAV/Plugin.php
@@ -77,6 +77,10 @@ class Plugin extends DAV\ServerPlugin
         // parents extend IExtendedCollection
         list($parent, $name) = Uri\split($uri);
 
+        if ('' === $uri) {
+            $parent = '';
+        }
+
         $node = $this->server->tree->getNodeForPath($parent);
 
         if ($node instanceof DAV\IExtendedCollection) {


evert:
I think it would be better to fix this where the function is called with `null`. It's better for functions to have a strict signature, and this feels like an bug

ByteHamster:
19f6055c0c5d68269c8ed5198ce5d43755ee4c80 ("Using strict_types everywhere") is the first bad commit. Fixing up the call stack could be done by:

-  Change `Uri\split($uri)` to return `""` instead of `null` if the regex does not match. Does not seem to be clean.
- Line `#2` in the stack trace seems to be ok. Getting methods for path `''` - no change needed. Line `#1` with `getNodeForPath(NULL)` looks bad but that line already is in `CalDAV/Plugin.php` and therefore probably does not fix the issue for all plugins. Would you still prefer to change in that location instead?

evert:
Strictly thinking about correct typing, I think it makes sense that `getNodeForPath` requires a string. After all, a path is always a string.

We _could_ update the method signature to say 'if NULL was given for a path, the root is returned', but it does open the door to potential unnoticed bugs.

So the 3 options really are:

1. Update getNodeForPath to treat `NULL` as an empty string.
2. As you said, always return a pair of strings from `Uri\split()`.
3. Whatever calls `split()` should handle the `null` case and transform it to a string.

Which approach is better? I'm not super sure.

One thought is that... given a path like `/foo`, `foo` clearly is the "basename" here, but what is the "dirname"? Is the parent path empty, or does the parent path not exist? This is somewhat a philosophical question? If there _is_ a parent path, and it's an empty string.. then what should `split('')` return? 

Anyway, those are the things bouncing around in my head. But thinking about this more, I don't have a real preference. I think they are all OK, and they all of trade-offs. I won't stand in the way either way.

ByteHamster:
> Update getNodeForPath to treat `NULL` as an empty string.

I agree with your first message: *It's better for functions to have a strict signature*. `NULL` is indeed not a path. So we should not accept it there.

> As you said, always return a pair of strings from `Uri\split()`.

We should probably not change this because the library is likely used by other projects, as well.

> Whatever calls `split()` should handle the `null` case and transform it to a string.

I changed the PR to do that. `Uri\split()` is called in many, many locations. Seems like it is enough to check in this single location though.

PR #1174, opened 2019-07-29
Add `afterCopy` event

### LINKED ISSUE #1076: No `afterCopy` event exists
No `afterCopy` event exists, similar to `afterMove`. This event is needed to copy properties over to the copied file.

fdoumet:
An afterCopy event is needed in order to perform custom actions such as clean up, or copying over custom properties similarly to afterMove. Fixes #1076

### DIFF
--- a/lib/DAV/PropertyStorage/Backend/BackendInterface.php
+++ b/lib/DAV/PropertyStorage/Backend/BackendInterface.php
@@ -77,4 +77,17 @@ interface BackendInterface {
      */
     function move($source, $destination);
 
+	/**
+	 * This method is called after a successful COPY
+	 *
+	 * This should be used to migrate all properties from one path to another.
+	 * Note that entire collections may be moved, so ensure that all properties
+	 * for children are also moved along.
+	 *
+	 * @param string $source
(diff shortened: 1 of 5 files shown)


lib/DAV/PropertyStorage/Backend/BackendInterface.php
```diff
@@ -77,4 +77,16 @@ function delete($path);
...
+	 *
+	 * This should be used to migrate all properties from one path to another.
+	 * Note that entire collections may be moved, so ensure that all properties
+	 * for children are also moved along.
+	 *
+	 * @param string $source
+	 * @param string $destination
+	 * @return void
+	 */
+    function copy($source, $destination);
```
staabm: I guess we need a new major version when changing this interface, because adding methods to it is a BC break

but lets see whats others think

evert: I agree. This is a hard BC break. Regardless I think new features should be avoided in old versions.

DeepDiver1975:
needs to target master branch - feel free to re-submit

PR #1175, opened 2019-08-06
fix Browser plugin open root path

asbiin:
In case a CalDAV uri exist, and you want to open the Browser plugin on root path, this error is thrown: 
```
Uncaught TypeError: trim() expects parameter 1 to be string, null given in sabre/dav/lib/DAV/Tree.php:57
```
This is caused by an empty value in `$uri`, and a `null` value in `$parent` variable.


DeepDiver1975:
duplicate of #1169

asbiin:
@DeepDiver1975 thanks!

PR #1203, opened 2019-10-22
Make it work when setBaseUri is set with on a folder and not a…

### LINKED ISSUE #1200: Plugin Browser : Guess Content Type doesn't work when setBaseUri is set with on a folder and not a php file
When the setBaseUri is set without a name file (without index.php for exemple) and directly on a folder, the script crash with the following error :

```
4.0.2 TypeError strrpos() expects parameter 1 to be string, null given
This XML file does not appear to have any style information associated with it. The document tree is shown below.
<s:sabredav-version>4.0.2</s:sabredav-version>
<s:exception>TypeError</s:exception>
<s:message>
strrpos() expects parameter 1 to be string, null given
</s:message>
<s:message>...</s:message>
</d:error>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">...</d:error>
```

I found the url more beauty without that and I achieve it with old version of sabre-dav.
I don't know whom file give me the error I don't have log of this error only this display.

When you enter a subdirectoy after the url, all work fine, but then you cannot acced with a standard client.

gloony:
Fix #1200

### DIFF
--- a/lib/DAV/Browser/GuessContentType.php
+++ b/lib/DAV/Browser/GuessContentType.php
@@ -85,12 +85,13 @@ class GuessContentType extends DAV\ServerPlugin
      */
     protected function getContentType($fileName)
     {
-        // Just grabbing the extension
-        $extension = strtolower(substr($fileName, strrpos($fileName, '.') + 1));
-        if (isset($this->extensionMap[$extension])) {
-            return $this->extensionMap[$extension];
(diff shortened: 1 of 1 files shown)


tchapi:
Hello ! Is this PR going to be merged soon ? I have the exact same problem and this PR is the perfect solution ;)

Thanks

PR #1201, opened 2019-10-21
Feature/streaming propfind

DeepDiver1975:
Using defibrillator on https://github.com/sabre-io/dav/pull/898

PR #1204, opened 2019-10-24
Feature/streaming propfind restart

DeepDiver1975:
Basically #1201 with clean git history and proper travis setup

## Description
- Server:generateMutliStatus can either return a callback or a string
- per default a string is returned in order to not break any existing functionality out there
- with Server::$streamMultiStatus the return can be controlled
- on travis the full test suite is executed with and without stream enabled

## ToDo
- [ ] run litmus

### DIFF
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -14,6 +14,7 @@ use Sabre\HTTP;
 use Sabre\HTTP\RequestInterface;
 use Sabre\HTTP\ResponseInterface;
 use Sabre\Uri;
+use Sabre\Xml\Writer;
 
 /**
  * Main DAV server class.
@@ -183,6 +184,15 @@ class Server implements LoggerAwareInterface, EmitterInterface
      * @var bool
      */
     public static $exposeVersion = true;
+
+    /**
+     * If this setting is turned on, any multi status response on any PROPFIND will be streamed to the output buffer.
+     * This will be beneficial for large result sets which will no longer consume a large amount of memory as well as
+     * send back data to the client earlier.
+     *
+     * @var bool
+     */
+    public static $streamMultiStatus = false;
 
     /**
      * Sets up the server.
@@ -1628,19 +1638,38 @@ class Server implements LoggerAwareInterface, EmitterInterface
     // {{{ XML Readers & Writers
 
     /**
-     * Generates a WebDAV propfind response body based on a list of nodes.
+     * Returns a callback generating a WebDAV propfind response body based on a list of nodes.
      *
      * If 'strip404s' is set to true, all 404 responses will be removed.
      *
      * @param array|\Traversable $fileProperties The list with nodes
      * @param bool               $strip404s
      *
-     * @return string
+     * @return callable|string
      */
     public function generateMultiStatus($fileProperties, $strip404s = false)
     {
         $w = $this->xml->getWriter();
+        if (self::$streamMultiStatus) {
+            return function () use ($fileProperties, $strip404s, $w) {
+                $w->openUri('php://output');
+                $this->writeMultiStatus($w, $fileProperties, $strip404s);
+                $w->flush();
+            };
+        }
         $w->openMemory();
+        $this->writeMultiStatus($w, $fileProperties, $strip404s);
+
+        return $w->outputMemory();
+    }
+
+    /**
+     * @param Writer $w
+     * @param $fileProperties
+     * @param bool $strip404s
+     */
+    private function writeMultiStatus(Writer $w, $fileProperties, bool $strip404s)
+    {
         $w->contextUri = $this->baseUri;
         $w->startDocument();
 
@@ -1662,7 +1691,6 @@ class Server implements LoggerAwareInterface, EmitterInterface
             ]);
         }
         $w->endElement();
-
-        return $w->outputMemory();
+        $w->endDocument();
     }
 }
--- a/tests/Sabre/CalDAV/ExpandEventsFloatingTimeTest.php
+++ b/tests/Sabre/CalDAV/ExpandEventsFloatingTimeTest.php
@@ -95,11 +95,12 @@ END:VCALENDAR
 
         $response = $this->request($request);
 
+        $bodyAsString = $response->getBodyAsString();
         // Everts super awesome xml parser.
         $body = substr(
-            $response->body,
-            $start = strpos($response->body, 'BEGIN:VCALENDAR'),
-            strpos($response->body, 'END:VCALENDAR') - $start + 13
+            $bodyAsString,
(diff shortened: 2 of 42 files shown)


lib/DAV/Server.php
```diff
@@ -561,7 +564,7 @@ public function calculateUri($uri)
             return trim(HTTP\decodePath(substr($uri, strlen($baseUri))), '/');
 
         // A special case, if the baseUri was accessed without a trailing
-        // slash, we'll accept it as well.
+            // slash, we'll accept it as well.
```
staabm:
```suggestion
        // slash, we'll accept it as well.
```

lib/DAV/Server.php
```diff
@@ -184,6 +185,8 @@ class Server implements LoggerAwareInterface, EmitterInterface
      */
     public static $exposeVersion = true;
 
+    public static $streamMultiStatus = false;
```
staabm: should have a proper phpdoc describing when/why it is usefull

lib/DAV/Server.php
```diff
@@ -1662,7 +1684,6 @@ public function generateMultiStatus($fileProperties, $strip404s = false)
             ]);
         }
         $w->endElement();
-
-        return $w->outputMemory();
+        $w->endDocument();
```
staabm: before we did not end the document? is this correct?
will the document always be generated using a single `generateMultiStatus()` call?

tests/Sabre/CalDAV/ExpandEventsFloatingTimeTest.php
```diff
@@ -95,11 +95,12 @@ public function testExpandCalendarQuery()
 
         $response = $this->request($request);
 
+        $bodyAsString = $response->getBodyAsString();
         // Everts super awesome xml parser.
         $body = substr(
-            $response->body,
-            $start = strpos($response->body, 'BEGIN:VCALENDAR'),
-            strpos($response->body, 'END:VCALENDAR') - $start + 13
+            $bodyAsString,
```
staabm: `$response->body` could be a string and a different type before, right?
we change it to a non-string with this PR just in even more places, correct? its not a new type for the body property?

tests/Sabre/CalDAV/Issue205Test.php
```diff
@@ -81,14 +81,14 @@ public function testIssue205()
 
         $response = $this->request($request);
 
-        $this->assertFalse(strpos($response->body, '<s:exception>Exception</s:exception>'), 'Exception occurred: '.$response->body);
-        $this->assertFalse(strpos($response->body, 'Unknown or bad format'), 'DateTime unknown format Exception: '.$response->body);
+        $this->assertFalse(strpos($response->getBodyAsString(), '<s:exception>Exception</s:exception>'), 'Exception occurred: '.$response->getBodyAsString());
```
staabm: do this multiple calls to `getBodyAsString()` have a perf impact?

DeepDiver1975:
litmus results:
```sh
✘ deepdiver:examples [ feature/streaming-propfind-restart ✭ | ✚ 1 …3 ⚑ 3 ] ➭ docker run -ti -e LITMUS_URL=http://172.17.0.1:6666/ -e LITMUS_USERNAME=user -e LITMUS_PASSWORD=123456 owncloud/litmus
-> running `basic':
 0. init.................. pass
 1. begin................. pass
 2. options............... pass
 3. put_get............... pass
 4. put_get_utf8_segment.. pass
 5. put_no_parent......... pass
 6. mkcol_over_plain...... pass
 7. delete................ pass
 8. delete_null........... pass
 9. delete_fragment....... pass
10. mkcol................. pass
11. mkcol_again........... pass
12. delete_coll........... pass
13. mkcol_no_parent....... pass
14. mkcol_with_body....... pass
15. finish................ pass
<- summary for `basic': of 16 tests run: 16 passed, 0 failed. 100.0%
-> running `copymove':
 0. init.................. pass
 1. begin................. pass
 2. copy_init............. pass
 3. copy_simple........... pass
 4. copy_overwrite........ pass
 5. copy_nodestcoll....... pass
 6. copy_cleanup.......... pass
 7. copy_coll............. pass
 8. copy_shallow.......... pass
 9. move.................. pass
10. move_coll............. pass
11. move_cleanup.......... pass
12. finish................ pass
<- summary for `copymove': of 13 tests run: 13 passed,

lib/DAV/Server.php
DeepDiver1975: pretty fix changed that - interesting .....

lib/DAV/Server.php
DeepDiver1975: for my understanding the endDocument() call was simply missing before and had no negative side effect. The writer is instanciated in generateMultiStatus and lives only within the scope of this method.

As far as I can tell this is correct.

tests/Sabre/CalDAV/ExpandEventsFloatingTimeTest.php
DeepDiver1975: $response->body can be of a different type - yes
Using getBodyAsString() all over the place in tests is basically the simplifying part to get the body always as string.
But getBodyAsString() can only be called once - this is why in many places a local variable is used to hold the string.

tests/Sabre/CalDAV/Issue205Test.php
DeepDiver1975: getBodyAsString() can only be called once in case the body is a resource or callable

DeepDiver1975:
> I assume this will be considered a BC breaking change and we will have a new major then?

My intention was by introducing the static property Server::$streamMultiStatus to avoid a BC.
Let me know if you have a different observation on this. THX

staabm:
> My intention was by introducing the static property Server::$streamMultiStatus to avoid a BC.

ACK. as long we are BC by default thats fine.

DeepDiver1975:
@evert Can I ask you for a second review? I'd love to get your feedback on this - THX

lib/DAV/Server.php
evert: Should this really be a static option?

evert:
Surprisingly simple PR, just a ton of test changes. Lgtm

lib/DAV/Server.php
DeepDiver1975: TBH: this was the simplest way to run the full test suite with minimal change 😬

DeepDiver1975:
If there are no further complains I suggest to put this into a 4.1.0-alpha. Objections?

staabm:
Do we expect perf/memory improvements caused by this change?

phil-davis:
Note: this ended up in the next release, that was numbered 4.0.3
So it is a "new optional feature" but a bit accidentally the release numbering is just a patch release.
Regardless of the semver numbering accident, it is BC anyway, so should not cause anyone grief.

Issue #1149, opened 2017-05-22
ErrorException Undefined offset: 1 in findByUri


muelli:
```
< 
< <?xml version="1.0" encoding="utf-8"?>
< <d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
<   <s:sabredav-version>3.1.3</s:sabredav-version>
<   <s:exception>ErrorException</s:exception>
<   <s:message>Undefined offset: 1</s:message>
< </d:error>
```

vendor/sabre/dav/lib/DAVACL/PrincipalBackend/PDO.php should use something like `array_pad`:

```
    function findByUri($uri, $principalPrefix) {
        $value = null;
        $scheme = null;
        list($scheme, $value) = array_pad(explode(":", $uri, 2), 2, null);
        if ($value == null) return null;

```

This is probably a sabredav issue though, but I don't know where to report it to.

staabm:
E.g. https://github.com/thephpleague/uri-parser

evert:
yeah or sabre/uri. I don't care either way but I think we already use that dependency ;)

staabm:
Ohh right.. didn‘t remember there is a sabre equivalent.

Sabre/uri it should be then

MTindev:
Issue linked to #1137

staabm:
PR welcome

tchapi:
Here you go : https://github.com/sabre-io/dav/pull/1210

staabm:
fixed with https://github.com/sabre-io/dav/pull/1210

PR #1210, opened 2019-11-19
Fix #1137 and #1149

### LINKED ISSUE #1137: PDO::findByUri() in error (when using Sharing with "principals" URI instead of e-mail address)
The method `Sabre\DAVACL\PrincipalBackend\PDO::findByUri($uri, $principalPrefix)` accepts as `$uri` parameter a string value, as documented _This method may receive any type of uri, but mailto: addresses will be the most common._

The [current implementation](https://github.com/sabre-io/dav/blob/master/lib/DAVACL/PrincipalBackend/PDO.php#L302-L335) takes into account two cases : a `mailto:ADDRESS` or a string which is simply returned. This last case allows the `$uri` string to be directly a `principals/NAME`.

However, the code beginning the implementation, [`list($scheme, $value) = explode(':', $uri, 2);`](https://github.com/sabre-io/dav/blob/master/lib/DAVACL/PrincipalBackend/PDO.php#L306), throws an error, in case of the `$uri` parameter which has not the format `mailto:ADDRESS`.

In fact `principals/NAME` can not be exploded in a 2-items array.
Error : **Notice: Undefined offset: 1**.

This error was found when using `Sabre\DAV\Client::shareCalendar()`. Probably the same via the interface calendar page sharing form (wish use `Sabre\DAV\Sharing\Plugin`)

tchapi:
Allow to find by principal uri
Fix uri parsing with sabre/uri

Added relevant phpunit tests

### DIFF
--- a/lib/DAVACL/PrincipalBackend/PDO.php
+++ b/lib/DAVACL/PrincipalBackend/PDO.php
@@ -301,19 +301,21 @@ class PDO extends AbstractBackend implements CreatePrincipalSupport
      */
     public function findByUri($uri, $principalPrefix)
     {
-        $value = null;
-        $scheme = null;
-        list($scheme, $value) = explode(':', $uri, 2);
-        if (empty($value)) {
+        $uriParts = Uri\parse($uri);
+
+        // Only two types of uri are supported :
+        //   - the "mailto:" scheme with some non-empty address
+        //   - a principals uri, in the form "principals/NAME"
+        // In both cases, `path` must not be empty.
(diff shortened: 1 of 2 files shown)


tchapi:
PS : your PrettyCI configuration / pipeline is not explained in the repo, and running it against the codebase with the `.php_cs.dist` file flags approximately _all_ the files of the repo as failed ... I only corrected my modifications then.

staabm:
thx for your contribution!

PR #1225, opened 2020-01-19
Bump dependencies

phil-davis:
1) Add .idea (PHPstorm IDE settings) to .gitignore so nobody accidentally commits it
2) Bump phpstan to current version 0.12.5 (the one that was hardcoded in `.travis.yml`)
3) remove duplicate bin/phpunit from .gitignore
4) switch to using phpstan from composer.json

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ services:
   - postgresql
 
 install:
-  - if [ $RUN_PHPSTAN == "TRUE"  ]; then wget https://github.com/phpstan/phpstan/releases/download/0.11.8/phpstan.phar; fi
+  - if [ $RUN_PHPSTAN == "TRUE"  ]; then composer require --dev phpstan/phpstan:^0.12; fi
 
 before_script:
 #  - mysql -u root -h 127.0.0.1 sabredav_test -e 'SELECT VERSION();'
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -30,7 +30,7 @@ services:
   - postgresql
 
 install:
-  - if [ $RUN_PHPSTAN == "TRUE"  ]; then wget https://github.com/phpstan/phpstan/releases/download/0.11.8/phpstan.phar; fi
+  - if [ $RUN_PHPSTAN == "TRUE"  ]; then wget https://github.com/phpstan/phpstan/releases/download/0.12.5/phpstan.phar; fi
```
staabm: In other projects I usually use composer to install pphstan with w ^0.12 requirement so it should auto update the minor releases

staabm: composer require --dev phpstan/phpstan:^0.12

phil-davis:
hmmm - `phpstan `0.12.*` requires PHP 7.1+ - I will think about what to do for that (actually we do not run `phpstan` on PHP 7.0, so the CI fail is "bogus" anyway)

composer.json
```diff
@@ -35,7 +35,8 @@
     "require-dev" : {
         "phpunit/phpunit" : "^6",
         "evert/phpdoc-md" : "~0.1.0",
-        "monolog/monolog": "^1.18"
+        "monolog/monolog": "^1.18",
+        "phpstan/phpstan": "^0.12"
```
staabm: Fyi, this is possible since phpstan 0.12 which no longer defines external dependencies

staabm:
> hmmm - `phpstan `0.12.*`requires PHP 7.1+ - I will think about what to do for that (actually we do not run`phpstan` on PHP 7.0, so the CI fail is "bogus" anyway)

Just make sure we ran phpstan on php 7.1+, thats fine.. doesnt make sense to make it compatible with 0.11

phil-davis:
The "problem" is that on PHP 7.0 Travis jobs, `composer update` "spits the dummy" if we specify `^0.12` (even though we do not care about `phpstan` at all on PHP 7.0)
I don't see a way in composer to specify that a dev dependency is only needed for PHP 7.1+

staabm:
Instead of adding it to the composer.json, just run the command in the build config:


composer require --dev phpstan/phpstan:^0.12

phil-davis:
I adjusted to require phpstan from Travis, squashed the commits.

staabm:
thx!

PR #1227, opened 2020-01-20
Drop PHP 7.0 and adjust tests for phpunit7

phil-davis:
I tried locally running phpunit7 and PHP 7.4.
This fixes things so that it passes.

1) `beStrictAboutTestSize` seems to have been deprecated and is not called `enforceTimeLimit`
2) when there are multiple `testSuite` entries in phpunit.xml they should be enclosed in `testsuites` 
3) A few phpunit Asserts have the message parameter as something other than a string - fix those
4) 1 failure happened:
```
There was 1 failure:

1) Sabre\CalDAV\Schedule\IMipPluginTest::testDeliverInsignificantRequest
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'1.0'
+'1'

/home/phil/git/sabre-io/dav/tests/Sabre/CalDAV/Schedule/IMipPluginTest.php:208
```
The result of the call to `getScheduleStatus()` is actually the string `1` and not `1.0` - I changed the the assert to `assertSame` so that this behaviour is not hidden by the test.

Is it supposed to return the string `1.0` ?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,10 +15,10 @@
     ],
     "require": {
         "php": "^7.1.0",
-        "sabre/vobject": "^4.2.0-alpha1",
+        "sabre/vobject": "^4.2.1",
         "sabre/event" : "^5.0",
         "sabre/xml"  : "^2.0.1",
-        "sabre/http" : "^5.0",
+        "sabre/http" : "^5.0.5",
         "sabre/uri" : "^2.0",
         "ext-dom": "*",
         "ext-pcre": "*",
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": ">=7.0.0",
+        "php": "^7.1.0",
         "sabre/vobject": "^4.2.0-alpha1",
         "sabre/event" : "^5.0",
         "sabre/xml"  : "^2.0.1",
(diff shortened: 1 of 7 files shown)


phil-davis:
Actually phpunit6 still seems to run OK on PHP 7.4 - but it seems "a good thing" to fix things that phpunit7 is reporting.

We will need to keep running phpunit6 as long as PHP 7.0 support is required. Or can we now drop PHP 7.0 support?

(and then there is dropping PHP 7.1 and moving to phpunit8...)

staabm:
IMO we should drop at least 7.0.

in case we see major problems, I am also ok with dropping 7.1.
in case its no major problem to support 7.1 we can/should keep it for a while.

phil-davis:
- dropped PHP 7.0 from tests
- bumped the PHP version in `composer.json` to 7.1
- bumped phpunit to v7

now I will look at the messages coming out of PHP7.,.4+phpunit7 test run and fix that stuff

composer.json
```diff
@@ -14,7 +14,7 @@
         }
     ],
     "require": {
-        "php": ">=7.0.0",
+        "php": ">=7.1.0",
```
staabm: we should make sure we dont collide with php8

```suggestion
        "php": "^7.1.0",
```

phil-davis:
@staabm Travis is green again. It needed min versions for `sabre/vobject` and `sabre/http` - earlier patch versions of those had messages that PHP7.4 reports.

covdecov fails marginally. I see that phpunit is running some tests from `/vendor/sabre/*` dependencies. I suppose if some stuff has change there then the codecov calculation could be a little different.

Ready for review.

staabm:
looks really good, thx

PR #1229, opened 2020-01-24
Server emails when calendar shared

danastasio:
Server will now send an email to users when a calendar is shared, or access is revoked.

### DIFF
--- a/lib/DAV/Sharing/Plugin.php
+++ b/lib/DAV/Sharing/Plugin.php
@@ -265,19 +265,21 @@ class Plugin extends ServerPlugin
             }
         }
 
-        $output .= '<tr><td colspan="2"><form method="post" action="">
-            <h3>Share this resource</h3>
-            <input type="hidden" name="sabreAction" value="share" />
-            <label>Share with (uri):</label> <input type="text" name="href" placeholder="mailto:user@example.org"/><br />
-            <label>Access</label>
-                <select name="access">
-                    <option value="readwrite">Read-write</option>
-                    <option value="read">Read-only</option>
-                    <option value="no-access">Revoke access</option>
-                </select><br />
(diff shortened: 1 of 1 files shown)


Tntdruid:
Very nice 👍

gruenie:
Because ByteHamster told us to discuss this subject here, I copy my last comment to here to continue from this point:
As I said. When I use DAVdroid (new name is DAVx5), then I see in my account all calendars (my own and the calendars which other people invited me to). There I can choose, which should be shown.
In Thunderbird (in te AddOn "Lightning") it is sadly completely different. There you need to add every calendar seperately (your own and others) by entering the completely url.
Thats not very nice, but at the moment it is like this. since Thunderbird is a really popular mail-client (for Windows and Linux), there should be a soltion too.

It is not the link that you wrote what we need to send in the mail, the link is:
https://server.tld/dav.php/calendars/Username/xxxxxxxx-yyyy-zzzz-xxxx-yyyyyyyyyyyy/,
where:
https://server.tld/dav.php/calendars/ - is the URL of the server to the calendars
Username - is the username of the receiver of the invitation (not the name of the sender)
xxxxxxxx-yyyy-zzzz-xxxx-yyyyyyyyyyyy - is the value, which you can find in the field uri in the table "calenderinstances".

I have no idea if the server-address is defined anywhere.
Maybe it is possible to get it from a $_SERVER-constant or what ever - otherwise it would be necessary to define it in a config-file.
More complicated will be to get the URI (calender-ID) from the database.

DeepDiver1975:
Sending email is a highly project specific feature and our of scope for this project. Please extend the functionality in our individual projects as needed. In case events are missing of private methods need to be protected to allow overloading feel free to submit pull requests. THX

PR #1232, opened 2020-01-28
Changed require_once -> Composer autoloader for test

C0pyR1ght:
Composer autoload_dev is now used for all Unittests. require_once was no longer neccessary and therefore removed.

resolves #1231

PR #1233, opened 2020-01-28
Implement php-cs-fixer in CI and apply changes

phil-davis:
PR #1232 introduced some empty lines that cause Pretty-CI to fail.

Add `php-cs-fixer` to the Travis CI (like in other sabre-io repos). It runs v2.16.1. Pretty-CI seems to be running 2.14.0, so it did not fuss about so much stuff.

`php-cs-fixer` found lots of params in PHPdoc to remove! (which already have their type specified in the function args) CI will tell us if it is OK.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,7 +25,6 @@ bin/phpdocmd
 bin/phpstan
 bin/phpstan.phar
 bin/php-cs-fixer
-bin/sabre-cs-fixer
 
 # Assuming every .php file in the root is for testing
 /*.php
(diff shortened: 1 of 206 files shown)


phil-davis:
@staabm are you "happy" that this will change 206 files?

staabm:
Thx for the PR. Didn‘t realize we dont have the travisci cs check running. 

I will remove prettyci for all sabre reposes shortly because it is stuck on cs-fixer  2.14 and will not update. 

Starting review now

PR #1234, opened 2020-01-29
rename phpunit.xml and standardize other tools

phil-davis:
Just a couple of little things to get the tooling/CI in this repo "standardized" like other repos.


bin/pretty
```diff
@@ -0,0 +1 @@
+../vendor/mnapoli/pretty/pretty
```
staabm: this shouldn't be here, right?

phil-davis: correct - random crud that got accidentally committed
removed it

staabm:
awesome, thx

PR #1238, opened 2020-02-03
WIP: migrate to PHPUnit8

C0pyR1ght:
- migrate Unittests
- update version in Composer.json

prerequisite

- sabre dependencies have to support PHPUnit 8

PR #1239, opened 2020-02-03
Phpunit8 migration

phil-davis:
- Commits from #1238 
- apply cs-fixer changes

Now to see what `phpunit` fails

Note: I will watch #1238 so I can pull any fixes/changes from there also.


phil-davis:
@staabm @C0pyR1ght this passes !!!

I will review the changes to see what else could be standardized or...

PR #1240, opened 2020-02-03
Standardize CI

phil-davis:
- cleanup `.gitignore`
- apply "standard" changes to `composer.json`
- fixup the stuff that `phpstan` and/or my IDE complains about in `tests`
(I deleted some helper code in `tests` where `phpstan` complained about something, but my IDE told me the helper method was unused - deleted code is debugged code!)


tests/Sabre/DAV/Mock/Collection.php
```diff
@@ -87,8 +87,11 @@ public function getName()
      *
      * @return string|null
      */
-    public function createFile($name, $data = '')
+    public function createFile($name, $data = null)
```
phil-davis: `phpstan` complained that in the parent class, `createFile` had a different signature ( `$data = null` )
So I made the signature match, then adjusted `$data` below so it will be OK to pass the empty string through to `new File()`

tests/Sabre/DAV/SimpleCollectionTest.php
```diff
@@ -44,6 +44,6 @@ public function testGetChild404()
     {
         $this->expectException('Sabre\DAV\Exception\NotFound');
         $s = new SimpleCollection('foo', []);
-        $this->assertEquals($s->getChild('404'));
+        $s->getChild('404');
```
phil-davis: This was a bogus call the `assertEquals()` (which should be given 2 parameters)
Actually we expect and exception, so we can just remove the `assertEquals`

tests/Sabre/DAVServerTest.php
```diff
@@ -193,7 +198,7 @@ public function initializeEverything()
     public function request($request, $expectedStatus = null)
     {
         if (is_array($request)) {
-            $request = HTTP\Request::createFromServerArray($request);
```
phil-davis: I have no idea what this was - `createFromServerArray` does not exist in `HTTP\Request`
I did find it in `HTTP\Sapi`

tests/Sabre/DAVServerTest.php
```diff
@@ -295,6 +300,6 @@ public function setUpBackends()
     public function assertHttpStatus($expectedStatus, HTTP\Request $req)
     {
         $resp = $this->request($req);
-        $this->assertEquals((int) $expectedStatus, (int) $resp->status, 'Incorrect HTTP status received: '.$resp->body);
+        $this->assertEquals((int) $expectedStatus, (int) $resp->getStatus(), 'Incorrect HTTP status received: '.$resp->getStatus());
```
phil-davis: I changed `$resp->body` in the message to just `$resp->getStatus()`
That should be enough, rather than dumping the whole body.

tests/bootstrap.php
```diff
@@ -44,6 +44,7 @@
 ];
 
 if (file_exists(__DIR__.'/config.user.php')) {
+    $userConfig = [];
```
phil-davis: `phpstan` complained that `$userConfig is undefined.
Initializing it here keeps `phpstan` quite, and also makes sure it has some value that would usually be overwritten by the following `include`

tests/phpunit.xml
```diff
@@ -7,7 +7,7 @@
   convertWarningsToExceptions="true"
   beStrictAboutTestsThatDoNotTestAnything="true"
   beStrictAboutOutputDuringTests="true"
-  enforceTimeLimit="true">
```
phil-davis: `enforceTimeLimit` needs `php-invoker`. We didn't use it in other repos.

staabm: > needs php-invoker

what does that mean? would it require a additional composer depdency?
(I am not used to this attribute)

phil-davis:
The previous PR does 3580 unit tests: https://travis-ci.org/sabre-io/dav/jobs/645495103?utm_medium=notification&utm_source=github_status

This PR only does 3736 unit tests: https://travis-ci.org/sabre-io/dav/jobs/645541039?utm_medium=notification&utm_source=github_status

ToDo: work out what I changed/deleted.

tests/phpunit.xml
phil-davis: I added `phpunit/php-invoker` to `composer.json` and put back `enforceTimeLimit`
IMO if a test "goes into a loop" then `phpunit` will be able to terminate it after some timeout.
The timeouts have reasonable defaults - I don't think we will need to specify them.

staabm: I see. not sure we need the additional complexity. I am not aware of a case in which our testsuite ran endless

staabm: I guess this is the only thing left... do we need the limit?

phil-davis: no, we don't need the limit. I will remove it.

phil-davis: removed

phil-davis:
@staabm ready for review again

.travis.yml
```diff
@@ -18,39 +18,37 @@ env:
...
 matrix:
   include:
     - name: 'PHPStan'
-      php: 7.2
+      php: 7.4
       env: RUN_PHPSTAN="TRUE"
     - name: 'Test with streaming propfind'
       php: 7.2
       env: RUN_TEST_WITH_STREAMING_PROPFIND="TRUE"
+  fast_finish: true
```
staabm: Did we use fast finish in other builds, too?

tests/Sabre/CalDAV/SharingPluginTest.php
```diff
@@ -328,7 +329,7 @@ public function testUnpublish()
         $request->setBody($xml);
 
         $response = $this->request($request);
-        $this->assertEquals(200, $response->status, $response->getBodyAsString());
+        $this->assertEquals(200, $response->getStatus(), $response->getBodyAsString());
```
staabm: Is there a reason why this property lookups are changed to getters?

.travis.yml
phil-davis: yes, it is like that in https://github.com/sabre-io/skel/blob/master/.travis.yml

tests/Sabre/CalDAV/TestUtil.php
```diff
@@ -99,87 +99,4 @@ public static function getTestCalendarData($type = 1)
 
         return $calendarData;
     }
-
-    public static function getTestTODO($type = 'due')
```
staabm: Unused method?

tests/Sabre/CardDAV/TestUtil.php
```diff
@@ -1,63 +0,0 @@
-<?php
-
-declare(strict_types=1);
-
-namespace Sabre\CardDAV;
-
-class TestUtil
```
staabm: Unused class?

tests/Sabre/CalDAV/SharingPluginTest.php
phil-davis: This was my IDE complaining "member has protected access".
I don't think `phpstan` complained about it.

staabm: I see

tests/Sabre/CalDAV/TestUtil.php
phil-davis: Yes, `phpstan` complained about stuff in it, and I noticed it was unused anyway.
```
Deleted code is quality code.
Deleted code is tested code.
```
:)

.travis.yml
staabm: Ok cool - didnt remember

tests/Sabre/CardDAV/TestUtil.php
phil-davis: Same here - `phpstan` had complaints. It was unused, so I deleted it.

tests/Sabre/CalDAV/SharingPluginTest.php
phil-davis: I can revert those, if you like.

staabm: Which IDE?

phil-davis: PHPstorm

phil-davis:
@staabm did you want me to make any changes? or is this OK?

tests/Sabre/CalDAV/SharingPluginTest.php
staabm: Only thing left is here: I was wondering why this changes are needed, or turned differently: why does phpstorm does not know about this properties.

can we do something about that?

phil-davis: I can revert these changes. I guess they are not necessary - the run-time PHP is able to access the properties.

phil-davis: Reverted

phil-davis:
@staabm unnecessary changes have been reverted

composer.json
```diff
@@ -70,7 +71,20 @@
         "bin/sabredav",
         "bin/naturalselection"
     ],
-    "config" : {
-        "bin-dir" : "./bin"
+    "scripts": {
+        "phpstan": [
+            "@php vendor/bin/phpstan analyse lib tests"
```
staabm: This should also work?

```suggestion
            "phpstan analyse lib tests"
```

composer.json
```diff
@@ -70,7 +71,20 @@
...
+        ],
+        "cs-fixer": [
+            "@php vendor/bin/php-cs-fixer fix"
+        ],
+        "phpunit": [
+            "@php vendor/bin/phpunit --configuration tests/phpunit.xml"
+        ],
+        "test": [
+            "composer phpstan",
```
staabm: Should work?

```suggestion
            "phpstan",
```

phil-davis: Doing that results in:
```
composer test
> phpstan
PHPStan - PHP Static Analysis Tool 0.12.9

Usage:
  command [options] [arguments]

Options:
  -h, --help            Display this help message
  -q, --quiet           Do not output any message
  -V, --version         Display this application version
      --ansi            Force ANSI output
      --no-ansi         Disable ANSI output
  -n, --no-interaction  Do not ask any interactive question
  -v|vv|vvv, --verbose  Increase the verbosity of messages: 1 for normal output, 2 for more verbose output and 3 for debug

Available commands:
  analyse    [analyze] Analyses source code
  dump-deps  Dumps files dependency tree
  help       Displays help for a command
  list       Lists commands
  worker     (Internal) Support for parallel analysis.
> composer cs-fixer
> @php vendor/bin/php-cs-fixer fix
...
```
It just does a "raw" `phpstan` command, rather than "looping back" into the `composer phpstan` to invoke the full command `phpstan analyse lib tests`

composer.json
phil-davis: Works. I simplified each of `phpstan` `cs-fixer` and `phpunit`
These need to be simplified in `skel` and the change applied across all repos.

phil-davis:
@staabm ready for review again.

Note: I simplified the `phpstan` command in Travis, because that exactly matches the `composer phpstan`  script that we already have. A bit of DRY.

For cs-fixer and phpunit, the commands in Travis use different parameters... so they cannot use the composer scripts that we have designed so far. But we could make extra composer script entries for `cs-fixer-dry-run` and the combinations of `phpunit` runs. That would get all the "magic different" tool commands in one place in `composer.json`. Is that worth doing?

composer.json
staabm: I see. When there is no perf difference in calling the process within a process within process I am fine with it

staabm: Feel free to merge

PR #1248, opened 2020-02-16
CalDAV to sync properly when limit is set in PDO backend

nhirokinet:
Hi,

I altered some codes to make CalDAV sync properly even when limit is set. Changes is mainly as below two:

- Return syncToken when limit is set. Current code returns the syncToken of the latest server state. Changed to return the syncToken corresponding to the truncated sync result.
  - When the result limit is designated by client, this is not the consistent as returned result.
  - In [RFC6578 section 3.6](https://tools.ietf.org/html/rfc6578#section-3.6): `the DAV:sync-token value returned in the response MUST represent the correct state for the partial set of
   changes returned.`
  - So, when truncation occurs, client will miss the truncated event even in later sync, since the client receives the syncToken of not received events, and asks server to send the event after that next time.
- When response is truncated, 507 record is added to the response.
  - Defined in [RFC6578 section 3.6](https://tools.ietf.org/html/rfc6578#section-3.6)
  - With this, the client can recognize the result is truncated, and can retry until the client syncs with the latest state.
  - For example, OneCalendar requested once I press sync button in current code. In new code OneCalendar repeated until it reaches the latest state, so I just had to press sync button once.

OneCalendar in Windows Store sends a request with limit 1000, and I checked using this, but with the changes below:

- I defanged the code of Authentication because OneCalendar did not send a request with Digest authentication.
- I overrided the limit to 2 in `lib/CalDAV/Backend/PDO.php` while testing.

I think CardDAV has the same issue, but I did not find an environment to reproduce and test with, so I wrote a code only for CalDAV, PDO backend.

This is my first code around WebDAV, so could you take a look including my understanding is correct or not?

### DIFF
--- a/tests/Sabre/DAV/Sync/PluginTest.php
+++ b/tests/Sabre/DAV/Sync/PluginTest.php
@@ -224,13 +224,18 @@ BLA;
         );
 
         $responses = $multiStatus->getResponses();
-        $this->assertEquals(1, count($responses), 'We expected exactly 1 {DAV:}response');
+        $this->assertEquals(2, count($responses), 'We expected exactly 2 {DAV:}responses');
 
         $response = $responses[0];
 
         $this->assertEquals('404', $response->getHttpStatus());
         $this->assertEquals('/coll/file3.txt', $response->getHref());
         $this->assertEquals([], $response->getResponseProperties());
+
+        $response = $responses[1];
+
+        $this->assertEquals('507', $response->getHttpStatus());
+        $this->assertEquals('/coll/', $response->getHref());
     }
 
     public function testSubsequentSyncSyncCollectionDepthFallBack()
--- a/lib/CalDAV/Backend/PDO.php
+++ b/lib/CalDAV/Backend/PDO.php
@@ -948,42 +948,46 @@ SQL;
         }
         list($calendarId, $instanceId) = $calendarId;
 
-        // Current synctoken
-        $stmt = $this->pdo->prepare('SELECT synctoken FROM '.$this->calendarTableName.' WHERE id = ?');
-        $stmt->execute([$calendarId]);
-        $currentToken = $stmt->fetchColumn(0);
-
-        if (is_null($currentToken)) {
-            return null;
-        }
-
         $result = [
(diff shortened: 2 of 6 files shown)


evert:
It's a bit hard for me to eye this and see if it's correct, but it *looks* good. I feel that this can use a good amount of unittests to ensure that edge cases are sufficiently covered and increase confidence.

nhirokinet:
I added test case for each of two changes.

nhirokinet:
@DeepDiver1975 I added some tests after your comments, should I add some more change?

nhirokinet:
Hi, is there something I can do to advance this pull request?
CI fails, but it seems to be some new timezone test, which this pull request does not touch.

nhirokinet:
@DeepDiver1975 Hi, do you have some comments about added tests?

nhirokinet:
@phil-davis @DeepDiver1975 Hi, do you know the status about this? I added some tests according to your comment, long ago, that I'm not sure about my memory, anything more I should do?

tests/Sabre/DAV/Sync/PluginTest.php
```diff
@@ -224,13 +224,18 @@ public function testSubsequentSyncSyncCollectionLimit()
         );
 
         $responses = $multiStatus->getResponses();
-        $this->assertEquals(1, count($responses), 'We expected exactly 1 {DAV:}response');
+        $this->assertEquals(2, count($responses), 'We expected exactly 1 {DAV:}response');
```
DeepDiver1975, line 227: > We expected exactly 1 {DAV:}response

We expected exactly 2 {DAV:}response

DeepDiver1975 (changes requested):

tests/Sabre/DAV/Sync/PluginTest.php
nhirokinet: fixed the message.

DeepDiver1975 (approved):

phil-davis (approved):

phil-davis:
Rebased and squashed. LGTM, thanks. Merging.

PR #1257, opened 2020-03-03
Fix #1254 - add missing MIME-Version header

### LINKED ISSUE #1254: Missing MIME-Version header in emails
When IMipPlugin is used to send emails for an event, it adds a `Content-Type` header, but forgets to add also a `MIME-Version` header which can trigger some spam engines

slaanesh:
When IMipPlugin is used to send emails for an event, it adds a Content-Type header, but forgets to add also a MIME-Version header which can trigger some spam engines. This PR fixes it and will address #1254 .


DeepDiver1975:
You need to adapt the tests as well - see https://travis-ci.org/sabre-io/dav/jobs/657685072#L435

THX

DeepDiver1975:
THX a lot!

PR #1265, opened 2020-04-22
Trigger CI

phil-davis:
to see what code style does


phil-davis:
With `--prefer-lowest` it passes. 
Without `--prefer-lowest` it fails. 

Something in new dependencies of `php-cs-fixer` is likely different. I will look.

PR #1267, opened 2020-04-22
Bump minimum php-cs-fixer version and adjust format of comment

### LINKED ISSUE #1266: php-cs-fixer CI is failing
e.g. see PR #1265 and #1245 

With `--prefer-lowest` it passes. 
Without `--prefer-lowest` it fails. 

Something in new dependencies of `php-cs-fixer` is likely different. I will look.

phil-davis:
Fixes #1266 

See issue  https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4933

Recent patch releases of `php-cs-fixer` have "fixed" the processing of comments at the top of the file (before any code) and now a 2nd comment is not considered part of the header. If the comment is about ordinary code (not about a variable) then the starting `/**` gets changes to `/*` (like it does if it occurs further down in the code)

`v2.16.2` onward behaves differently to `v2.16.1`. There is now `v2.16.3` so require at least a minimum of that `php-cs-fixer` patch level. And apply the comment change that it wants.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "ext-json": "*"
     },
     "require-dev" : {
-        "friendsofphp/php-cs-fixer": "~2.16.1",
+        "friendsofphp/php-cs-fixer": "^2.16.3",
         "phpstan/phpstan": "^0.12",
         "phpunit/phpunit" : "^7.5 || ^8.5 || ^9.0",
         "evert/phpdoc-md" : "~0.1.0",
--- a/examples/groupwareserver.php
+++ b/examples/groupwareserver.php
@@ -12,7 +12,7 @@
  * statement.
  */
 
-/**
+/*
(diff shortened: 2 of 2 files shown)


staabm:
Thanks for the investigation.. will wait what the upstream issue will provide

phil-davis:
I pinged upstream just now - let's see if any response.

phil-davis:
@staabm there was some discussion in the upstream issue https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4933 but it seems to have stopped getting more comments.

IMO the new behaviour is fine, and we should get CI green here.

Merge?

staabm:
I dont think the comment style within the example file looks consistent.

As its just a minor detail feel free to merge. I dont think we should spent too many time for it

Issue #1270, opened 2020-05-20
DAVx5: Received non-XML 207 Multi-Status


das-bub:
Hi,

I've already reported this [downstream](https://github.com/tchapi/davis/issues/8) and was told that it's probably a sabre/dav issue, so here it goes.

DAVx5 expects a Multi-Status response with the content-type of text/xml or application/xml, while sabre/dav responds with text/html, which breaks DAVx5.

According to everything I've read DAVx5 was working with sabre/dav before, so I'm really not sure what changed.

I'm using nginx 1.16 with php-fpm 7.3, davis 1.4.7 with sabre/dav 4.1.0 and DAVx5 3.0-ose2 (300000004).

As mentioned in the downsteram bug other clients, like iOS and the Thunderbird Lightning extension work as is.
I guess they just ignore the content-type.

Debug log: [davx5-info.txt](https://github.com/sabre-io/dav/files/4655458/davx5-info.txt)

DeepDiver1975:
sabre/dav is used without issues with the davx client in many projects (owncloud, nextcloud, baikal and many more).
Therefor I suspect the issue more within the downstream project and I kindly ask downstream to analyse the situation first within the project scope.
As soon as you are facing a real issue in sabre I'll happily help out. THX a lot

tchapi:
Hi Thomas — downstream project maintainer here 👋 

After investigation this specific issue is indeed on our side, but I think there is a wider problem that _could_ be addressed in **sabre/dav** to mitigate it, and it's basically that the `Server` classes uses the hardcoded `Sabre\HTTP\Sapi` class.

When using another framework that has its own and proven implementation of PSR-7, it's a hacky thing to get it to encapsulate the DAV server response since it directly writes to `php://output`.. and sends the headers right away.

It's basically the point of https://github.com/sabre-io/dav/issues/1037 — and it would be great to have some control over _how_ **sabre/dav** sends the response (could be great to have a `getResponse()` method to retrieve the full response object, so that we could send it ourselves, for instance) even if you do not go down the path of PSR-7.

In the meantime, I have fixed it (https://github.com/tchapi/davis/commit/a3dab81b4e25ba8ac146da6880974829e308dd65) by looking at the _prepared_ headers from **sabre/dav**. I guess it will do, but it would be much better to have the ability to do something before the headers are sent.

Happy to hear your thoughts whenever you have some time
Best !
Cyril

PS: @das-bub you can close this issue

das-bub:
[...]

PR #1271, opened 2020-05-22
Allow to use custom SAPI implementations - helpful when integrating w…

DeepDiver1975:
…ith different frameworks

refs #1270

### DIFF
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -208,8 +208,10 @@ class Server implements LoggerAwareInterface, EmitterInterface
      * the nodes in the array as top-level children.
      *
      * @param Tree|INode|array|null $treeOrNode The tree object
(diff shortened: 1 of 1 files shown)


lib/DAV/Server.php
```diff
@@ -208,8 +208,11 @@ class Server implements LoggerAwareInterface, EmitterInterface
      * the nodes in the array as top-level children.
      *
      * @param Tree|INode|array|null $treeOrNode The tree object
+     * @param HTTP\Sapi|null        $sapi
+     *
+     * @throws Exception
      */
-    public function __construct($treeOrNode = null)
+    public function __construct($treeOrNode = null, $sapi = null)
```
staabm, line 215: I think we should type this to `HTTP\Sapi`

PR #1273, opened 2020-05-29
The baseUri was missing when generating the lock responses. This was …

DeepDiver1975:
…leading to wrong responses.

### DIFF
--- a/tests/Sabre/DAV/Locks/PluginTest.php
+++ b/tests/Sabre/DAV/Locks/PluginTest.php
@@ -110,6 +110,32 @@ class PluginTest extends DAV\AbstractServer
         $this->assertEquals($this->response->getHeader('Lock-Token'), '<'.(string) $token[0].'>', 'Token in response body didn\'t match token in response header.');
     }
 
+    public function testLockWithContext()
+    {
+        $request = new HTTP\Request('LOCK', '/baseuri/test.txt');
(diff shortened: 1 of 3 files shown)


tests/Sabre/DAV/Locks/PluginTest.php
```diff
@@ -23,29 +23,29 @@ public function setup(): void
...
+    public function testGetInfo(): void
     {
         $this->assertArrayHasKey(
             'name',
             $this->locksPlugin->getPluginInfo()
         );
     }
 
-    public function testGetFeatures()
+    public function testGetFeatures(): void
```
staabm, line 34: could we separate this test-signature method related changes into a separate PR?

DeepDiver1975: I don't consider these changes critical - but if you like I remove this.

staabm: I would prefer a commit in which we can see the actual fix you did.

atm it feels the PR is 90% noise.

it would also be ok to have it in separate commits

DeepDiver1975: here we go .... @staabm

staabm:
Lgtm. Feel free to merge in case you verified the test breaks wirhout your patch

Thanks!

PR #1281, opened 2020-07-13
Release 4.1.1

phil-davis:
Includes:
#1269 Tree::move method returns no int
#1271 Allow to use custom SAPI implementations
#1273 Include baseUri in lock responses

And CI will run with the newly-released https://github.com/sabre-io/vobject/releases/tag/4.3.1 and demonstrate that the unit test warning has gone.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 ChangeLog
 =========
 
+4.1.1 (2020-07-13)
+-------------------------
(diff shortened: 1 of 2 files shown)


phil-davis:
https://travis-ci.org/github/sabre-io/dav/jobs/707595886
PHPUnit9 is passing without any warning or deprecation messages - good.

PR #1285, opened 2020-07-15
RFC 4918 compliance: Remove unexpected propstat

mstilkerich:
According to RFC 4918, there is two types of response elements:

See section 14.24, response XML Element:
```
<!ELEMENT response (href, ((href*, status)|(propstat+)),
                       error?, responsedescription? , location?) >
```

Type 1 contains one or more href and one status child elements.

Type 2 contains one href and one or more propstat child elements.

Both types may contain further optional elements.

For Type 1, sabre/dav inserts a spurious propstat element, which violated the element definition by RFC 4918.

You can also see examples for such type 1 response elements in RFC 4918, for example 9.6.2 example:

```
>>Request
DELETE /container/ HTTP/1.1 Host: www.example.com

>>Response
HTTP/1.1 207 Multi-Status
Content-Type: application/xml; charset="utf-8" Content-Length: xxxx
<?xml version="1.0" encoding="utf-8" ?> <d:multistatus xmlns:d="DAV:">
<d:response> <d:href>http://www.example.com/container/resource3</d:href> <d:status>HTTP/1.1 423 Locked</d:status> <d:error><d:lock-token-submitted/></d:error>
       </d:response>
     </d:multistatus>
```

This issue may cause failure with clients that validate the received XML against the RFC specification.

### DIFF
--- a/lib/DAV/Xml/Element/Response.php
+++ b/lib/DAV/Xml/Element/Response.php
@@ -112,12 +112,14 @@ class Response implements Element
      */
     public function xmlSerialize(Writer $writer)
     {
+        $empty = true;
+
         if ($status = $this->getHTTPStatus()) {
+            $empty = false;
             $writer->writeElement('{DAV:}status', 'HTTP/1.1 '.$status.' '.\Sabre\HTTP\Response::$statusCodes[$status]);
         }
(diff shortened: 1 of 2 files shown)


mstilkerich:
No one interested in fixing this?

alecpl (changes requested):

Issue #1286, opened 2020-07-22
Why isn't sabredav using the mime_content_type() php function?


AvverbioPronome:
https://github.com/sabre-io/dav/blob/5736f943c90d8d73d04cd8944d8c913811dc7360/lib/DAV/FS/File.php#L83

https://www.php.net/manual/en/function.mime-content-type.php

DeepDiver1975:
PR welcome - THX

AvverbioPronome:
I have a first iteration, I run the unit tests and I get 

```
Tests: 1605, Assertions: 1978, Errors: 306, Skipped: 219.
```

Is this expected?

Also: is there a dockerized test environment I can run?

DeepDiver1975:
Please submit the pr and we can have a look together

evert:
See discussion in #1287

PR #1287, opened 2020-07-22
use builtin php function instead of null default for determining mime…

AvverbioPronome:
see issue #1286

### DIFF
--- a/lib/DAV/FS/File.php
+++ b/lib/DAV/FS/File.php
@@ -82,6 +82,6 @@ class File extends Node implements DAV\IFile
      */
     public function getContentType()
     {
-        return null;
+        return mime_content_type($this->path);
     }
 }
--- a/lib/DAV/FS/File.php
+++ b/lib/DAV/FS/File.php
@@ -82,6 +82,6 @@ class File extends Node implements DAV\IFile
      */
     public function getContentType()
     {
-        return mime_content_type($this->path);
+        return @mime_content_type($this->path) ?: null;
     }
 }
--- a/tests/Sabre/DAV/Browser/GuessContentTypeTest.php
+++ b/tests/Sabre/DAV/Browser/GuessContentTypeTest.php
@@ -29,8 +29,8 @@ class GuessContentTypeTest extends DAV\AbstractServer
         ];
         $result = $this->server->getPropertiesForPath('/somefile.jpg', $properties);
         $this->assertArrayHasKey(0, $result);
-        $this->assertArrayHasKey(404, $result[0]);
-        $this->assertArrayHasKey('{DAV:}getcontenttype', $result[0][404]);
+        $this->assertArrayHasKey(200, $result[0]);
+        $this->assertArrayHasKey('{DAV:}getcontenttype', $result[0][200]);
     }
 
     /**
--- a/tests/Sabre/DAV/Browser/GuessContentTypeTest.php
+++ b/tests/Sabre/DAV/Browser/GuessContentTypeTest.php
@@ -12,7 +12,7 @@ class GuessContentTypeTest extends DAV\AbstractServer
     {
         parent::setUp();
         \Sabre\TestUtil::clearTempDir();
-        imagejpeg(imagecreatetruecolor(100,100), SABRE_TEMPDIR.'/somefile.jpg');
+        imagejpeg(imagecreatetruecolor(100, 100), SABRE_TEMPDIR.'/somefile.jpg');
         file_put_contents(SABRE_TEMPDIR.'/somefile.hoi', 'blabla');
     }
 
--- a/tests/Sabre/DAV/Browser/GuessContentTypeTest.php
+++ b/tests/Sabre/DAV/Browser/GuessContentTypeTest.php
@@ -12,7 +12,7 @@ class GuessContentTypeTest extends DAV\AbstractServer
     {
         parent::setUp();
         \Sabre\TestUtil::clearTempDir();
-        file_put_contents(SABRE_TEMPDIR.'/somefile.jpg', 'blabla');
+        imagejpeg(imagecreatetruecolor(100,100), SABRE_TEMPDIR.'/somefile.jpg');
         file_put_contents(SABRE_TEMPDIR.'/somefile.hoi', 'blabla');
     }
 
--- a/tests/Sabre/DAV/Browser/GuessContentTypeTest.php
+++ b/tests/Sabre/DAV/Browser/GuessContentTypeTest.php
@@ -13,7 +13,7 @@ class GuessContentTypeTest extends DAV\AbstractServer
         parent::setUp();
         \Sabre\TestUtil::clearTempDir();
         imagejpeg(imagecreatetruecolor(100, 100), SABRE_TEMPDIR.'/somefile.jpg');
-        file_put_contents(SABRE_TEMPDIR.'/somefile.hoi', 'blabla');
+        file_put_contents(SABRE_TEMPDIR.'/somefile.hoi', random_bytes(1024));
     }
 
     public function tearDown(): void
--- a/tests/Sabre/DAV/ServerSimpleTest.php
+++ b/tests/Sabre/DAV/ServerSimpleTest.php
@@ -98,7 +98,7 @@ class ServerSimpleTest extends AbstractServer
 
         $this->assertEquals([
             'X-Sabre-Version' => [Version::VERSION],
-            'Content-Type' => ['application/octet-stream'],
+            'Content-Type' => [mime_content_type($filename)], // or should we hardcode text/plain?
             'Content-Length' => [13],
             'Last-Modified' => [HTTP\toDate(new \DateTime('@'.filemtime($filename)))],
             'ETag' => ['"'.sha1(fileinode($filename).filesize($filename).filemtime($filename)).'"'],
--- a/tests/Sabre/DAV/ServerSimpleTest.php
+++ b/tests/Sabre/DAV/ServerSimpleTest.php
@@ -98,7 +98,7 @@ class ServerSimpleTest extends AbstractServer
 
         $this->assertEquals([
             'X-Sabre-Version' => [Version::VERSION],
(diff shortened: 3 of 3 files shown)


DeepDiver1975:
Only two failures - related to the change:

```
There were 2 failures:

1) Sabre\DAV\Browser\GuessContentTypeTest::testGetProperties

Failed asserting that an array has the key '{DAV:}getcontenttype'.

/home/travis/build/sabre-io/dav/tests/Sabre/DAV/Browser/GuessContentTypeTest.php:33

2) Sabre\DAV\ServerSimpleTest::testBaseUri

Failed asserting that two arrays are equal.

--- Expected

+++ Actual

@@ @@

 Array (

     'X-Sabre-Version' => Array (...)

     'Content-Type' => Array (

-        0 => 'application/octet-stream'

+        0 => 'text/plain'

     )

     'Content-Length' => Array (...)

     'Last-Modified' => Array (...)

     'ETag' => Array (...)

 )

/home/travis/build/sabre-io/dav/tests/Sabre/DAV/ServerSimpleTest.php:106

**Comment by AvverbioPronome** [comment]
The first I don't understand how could happen

The second just looks like php has a different default mimetype.

**Comment by DeepDiver1975** [comment]
> The second just looks like php has a different default mimetype.

no - in case the mime type cannot be determined - the code base will apply application/octet as default - with your change this is now different

**Comment by DeepDiver1975** [comment]
> The first I don't understand how could happen

https://github.com/sabre-io/dav/blob/95bdc9259099fe1d69955c8cccf75fdb641552e9/tests/Sabre/DAV/Browser/GuessContentTypeTest.php#L33
before your fix that property was in the 404 result set which is no longer is now

**Comment by AvverbioPronome** [comment]
Oh, ok. Should I remove line 33 from that test and update the mimetype for the other?

**Comment by DeepDiver1975** [comment]
It should look like this (not tested - just a brain dump)

```
$this->assertArrayHasKey('{DAV:}getcontenttype', $result[0][200]);

lib/DAV/FS/File.php
```diff
@@ -82,6 +82,6 @@ public function getETag()
      */
     public function getContentType()
     {
-        return null;
+        return mime_content_type($this->path);
```
staabm, line 85: in case `mime_content_type` returns `false` because of errors, should we return `null` here?

AvverbioPronome: it should probably be handled as an exception, `mime_content_type` goes to `E_WARNING` on failure. -- and it makes sense, such failure happens when the os is unable to determine the mimetype, and it sounds unusual.

AvverbioPronome:
> no - in case the mime type cannot be determined - the code base will apply application/octet as default - with your change this is now different

This is not the case, I think. If [this](https://github.com/sabre-io/dav/blob/5736f943c90d8d73d04cd8944d8c913811dc7360/tests/Sabre/DAV/ServerSimpleTest.php#L110) is the right test I'm triggering, an ascii string should not get identified as `application/octet-stream` instead of `text/plain`.

So, if `test.txt` gets served as `text/plain` that's a desirable outcome, imho.

AvverbioPronome:
I believe this modifications are ready to be tested. There are two open questions:

What to do here:

https://github.com/AvverbioPronome/dav/blob/e5abbd52f71ea1bbd7eea6f73ca46deb3a6221a2/tests/Sabre/DAV/ServerSimpleTest.php#L101

And what to do about the `GuessContentType` plugin, which gets obsoleted by there changes.

lib/DAV/FS/File.php
staabm: I still think we should return null then, because the caller code is already able to handle that case.

tests/Sabre/DAV/ServerSimpleTest.php
```diff
@@ -98,7 +98,7 @@ public function testBaseUri()
 
         $this->assertEquals([
             'X-Sabre-Version' => [Version::VERSION],
-            'Content-Type' => ['application/octet-stream'],
+            'Content-Type' => [mime_content_type($filename)], // or should we hardcode text/plain?
```
staabm, line 101: in a test-case we should hard-code the value

tests/Sabre/DAV/Browser/GuessContentTypeTest.php
```diff
@@ -29,8 +29,8 @@ public function testGetProperties()
         ];
         $result = $this->server->getPropertiesForPath('/somefile.jpg', $properties);
         $this->assertArrayHasKey(0, $result);
-        $this->assertArrayHasKey(404, $result[0]);
-        $this->assertArrayHasKey('{DAV:}getcontenttype', $result[0][404]);
+        $this->assertArrayHasKey(200, $result[0]);
```
staabm, line 32: do we have a test-case with a non existant file, which returns http 404 for a non existant file?

I assume the test here will find `somefile.jpg` and therefore returns 200?

AvverbioPronome: I don't know. But it's probably outside the scope of GuessContentType. 

Anyhow, the [following test](https://github.com/sabre-io/dav/blob/e5abbd52f71ea1bbd7eea6f73ca46deb3a6221a2/tests/Sabre/DAV/Browser/GuessContentTypeTest.php#L39) **will** fail as well because `somefile.jpg`, despite the extension,[ is not ](https://github.com/sabre-io/dav/blob/e5abbd52f71ea1bbd7eea6f73ca46deb3a6221a2/tests/Sabre/DAV/Browser/GuessContentTypeTest.php#L15) an `image/jpeg` file. But the test makes sense if you're testing GuessContentType, which determines mime type by looking at the file extension.

Instead, `mime_content_type` uses the [same mechanism as the `file` command](https://linux.die.net/man/1/file).

There is no reason to keep the plugin and its tests, imho.

tests/Sabre/DAV/ServerSimpleTest.php
AvverbioPronome: done.

lib/DAV/FS/File.php
AvverbioPronome: done this as well.

evert:
Hi!

So, at least one reason for GuessContentType's pretty small list of mimetypes, is that letting everything through opens the door to potential security issues.

For instance, I can upload a html file and add some javascript. Next user comes along and starts executing. It would basically be a pretty big XSS issue / complete account take over.

It seemed prudent at the time to use a pretty conservative list for this, and only let things through we know are low risk.

Other than the (small?) additional overhead I don't really see any issues though, but I feel that this needs to at least be addressed.

AvverbioPronome:
> Next user comes along and starts executing.

Yeah, that should not happen.

But I think the solution to that problem is to use [`Content-Disposition: attachment`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) when needed, not to make the `Content-Type` header useless.

evert:
Forcing a `Content-Disposition: attachment` might also have unintended consequences. People might actually intentionally directly embed things. This is definitely a use-case I've seen.

I think the safest is to stick with a strict list of content-types that are explicitly allowed. Let users decide what they want to support. I don't mind expanding the current list with more (safe) mimetypes.

Of course there is nothing preventing you from subclassing `FS\File` and set the behavior that works for you. sabre/dav was always a framework for building DAV applications first, with reasonable default implementations.If you need features that are non-default, there are extension points via event handlers and custom nodes

AvverbioPronome:
> reasonable

Yeah... I don't think it's very reasonable to use `Content-Type` this way. And as you said using `Content-Disposition` might have unintended consequences as well. 

I think an alternative can be to have sabre/dav know internally the actual mimetypes, safe or unsafe, and then eventually apply a user-configurable whitelist at the very last moment, in [this function](https://github.com/sabre-io/dav/blob/5736f943c90d8d73d04cd8944d8c913811dc7360/lib/DAV/Server.php#L849), just before serving the file.

evert:
Adding a central filter for content-types is *also* a BC break because currently anyone can still define their own mimetypes in their own nodes.

I'm not convinced this is useful enough, but I'll leave it up to @staabm and @DeepDiver1975 what they want to do with this.

AvverbioPronome:
@evert there's yet another compromise: don't set a mimetype, let the webserver handle it. 

AFAIK there are a lot of use cases for different users. I'm probably thinking too much about the needs of users of private shares only.

evert:
I'm not sure webservers will autodetect based on what stream PHP returns. But if this works for you, great!

staabm:
> I'm not sure webservers will autodetect based on what stream PHP returns.

I dont think this can work. Webserver only add this types based in filenames/file-extensions or directory-name

AvverbioPronome:
Should I convert this pull request into a plugin?

AvverbioPronome:
> AFAIK Webserver only add this types based in filenames/file-extensions or directory-name

And this should be fine (even though I agree it's not perfect), the point of me suggesting that is to move the mimetype-security burden from sabre/dav to the webserver, which I can assume the user has configured according to their liking.

evert:
> And this should be fine (even though I agree it's not perfect), the point of me suggesting that is to move the mimetype-security burden from sabre/dav to the webserver, which I can assume the user has configured according to their liking.

No, this is just not how webservers work for dynamic content, or how they are generally configured. Consider the `text/html` case. By default any webserver will allow text/html to be generated by files or dynamic endpoints such as PHP. You usually want PHP to emit HTML after all.

Webservers generally trust PHP or any files in the filesystem, because webservers assume that only people with appropriate access will be able to create these files. This is different when you deal with end-users.

evert:
> Should I convert this pull request into a plugin?

Plugins are also a pain because you don't necessarily operate on files. What if the backend store was S3, or a database? These are common use-cases. If we do a mechanism based on magic bytes, it would mean we have to read streams and potentially rewind them, which is *also* not always possible.

I stand by my original point. If you have a need to do this, sabre/dav already provides all the extension points for you to do so. This doesn't need to be a core thing for all the above reasons.

So based on all this, I'm going to close this ticket. I do appreciate the contribution and hope you are not discouraged from making further PR's and suggestions.

PR #1290, opened 2020-08-11
add params for put interface

yrong:
for later usage like https://github.com/nextcloud/server/issues/22019


phil-davis:
@yrong I made a rebased PR  #1365 and added PHPdoc etc there.

Is this enhancement still required?

phil-davis:
Merged in #1365

phil-davis:
@yrong see discussion in https://github.com/sabre-io/dav/pull/1365#discussion_r765559132

This change, as it is, is not strictly backward-compatible. I had released it today in 4.2.2. I had to revert it and release 4.2.3.

If you need this change/enhancement then please mention here the detail of where it will be used. Then we can think how best to add it.

PR #1292, opened 2020-08-11
Fix principal search to handle inaccessible principals

### LINKED ISSUE #1291: REPORT throws TypeError if principalSearch finds principals that user has no read permission on
If I search for principals when adding event attendees and that search matches principals that I don't have read access too, the server returns an error.

`TypeError "ltrim() expects parameter 1 to be string, null given" thrown in ...vendor/sabre/dav/lib/DAV/Server.php on line 1663`

## Steps to reproduce
1. Add custom principals for your PrincipalCollection (see Environment) to server structure.
2. Create two separate groups in db.
3. Create multiple principals in db with similar display names and assign them to one of the groups.
4. Setup one of the accounts in OSX Calendar.
5. Create a new Event.
6. Start adding new Attendees: Let the search lookup possible attendees.
7. Notice that nothing will be returned if search letters matches display name of Principal that is not part of the user’s group.

**Example Request**
Request Url: `REPORT http://john.doe%40example.com@0.0.0.0:8080/principals/`
```xml
<?xml version="1.0" encoding="UTF-8"?>
<A:principal-property-search xmlns:A="DAV:" test="anyof">
  <A:property-search>
    <A:prop>
      <A:displayname/>
    </A:prop>
    <A:match match-type="starts-with">jan</A:match>
  </A:property-search>
  <A:prop>
    <B:calendar-user-type xmlns:B="urn:ietf:params:xml:ns:caldav"/>
    <B:calendar-user-address-set xmlns:B="urn:ietf:params:xml:ns:caldav"/>
    <C:record-type xmlns:C="http://calendarserver.org/ns/"/>
    <A:principal-URL/>
    <A:displayname/>
  </A:prop>
</A:principal-property-search>
```

### Expected Result
The server returns all matched principals that i have read access to.

### Actual Result
The server returns an error.

**Response**
```xml
  <s:sabredav-version>4.1.1</s:sabredav-version>
  <s:exception>TypeError</s:exception>
  <s:message>ltrim() expects parameter 1 to be string, null given</s:message>
</d:error>
```

## Description
I think the problem lies within the DAVACL plugin. The *principalSearch* method does not handle when nothing is returned from *$this->server->getPropertiesForPath*.
The principalSearch first gets all matching principals from the principal backend. This also contains principals that the current user has no read rights on since this check does not happen at this point. Afterwards the method wants to populate the found principals with the requested properties (*$this->server->getPropertiesForPath*). Here ACL gets checked via the *propFind* Event and the method returns no properties if the current user has no read access to that principal.
This is already causing this notice to appear:
`PHP Notice:  Undefined offset: 0 in ...vendor/sabre/dav/lib/DAVACL/Plugin.php on line 754`
With that the *principalSearch* method returns an array that can contain null values.
Now the original caller method *principalPropertySearchReport* calls *$this->server->generateMultiStatus* with an array with null values which makes the server throw an TypeError.
`TypeError "ltrim() expects parameter 1 to be string, null given" thrown in ...vendor/sabre/dav/lib/DAV/Server.php on line 1663`

## Environment
PHP 7.2
sabre/dav 4.1.1
OSX Calendar

I handle visiblity of Principal nodes in a way that only allows members of the same group to see each other.

```php
public function getACL()
    {
        $acl = parent::getACL();
        $groups = $this->getGroupMembership();
        foreach ($groups as $group) {
            $acl[] = [
                'privilege' => '{DAV:}read',
                'principal' => $group,
                'protected' => true,
            ];
        }
        return $acl;
    }
```
*getACL method of my custom Principal Class (extends CalDav\Principal\User)*

To use my Principal class, I also have my own Collection class that extends the normal CalDAV\Principal\Collection and returns objects of my custom Principal class.

Additionally, Errors and warnings get converted to Exceptions (Including the type error).

mergeMarc:
Fixes #1291 by checking if property lookup returns an empty array and if so, not including that principal in the search results.

### DIFF
--- a/lib/DAVACL/Plugin.php
+++ b/lib/DAVACL/Plugin.php
@@ -751,7 +751,12 @@ class Plugin extends DAV\ServerPlugin
         $matches = [];
 
         foreach ($lookupResults as $lookupResult) {
-            list($matches[]) = $this->server->getPropertiesForPath($lookupResult, $requestedProperties, 0);
+            $properties = $this->server->getPropertiesForPath($lookupResult, $requestedProperties, 0);
+            if (empty($properties)) {
+                // Ignore found principals that we cannot access
+                continue;
+            }
+            list($matches[]) = $properties;
         }
 
         return $matches;
--- a/tests/Sabre/DAVACL/PrincipalPropertySearchBlockTest.php
+++ b/tests/Sabre/DAVACL/PrincipalPropertySearchBlockTest.php
@@ -1,0 +1,323 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabre\DAVACL;
+
+use Sabre\DAV;
+use Sabre\HTTP;
+
+class PrincipalPropertySearchBlockTest extends \PHPUnit\Framework\TestCase
+{
+    public function getServer()
+    {
+        $backend = new PrincipalBackend\Mock();
+
+        $dir = new DAV\SimpleCollection('root');
+        $principals = new PrincipalCollection($backend);
+        $dir->addChild($principals);
+
+        $fakeServer = new DAV\Server($dir);
+        $fakeServer->sapi = new HTTP\SapiMock();
+        $fakeServer->httpResponse = new HTTP\ResponseMock();
+        $fakeServer->debugExceptions = true;
+        $plugin = new MockPluginRestrictive();
+        $plugin->allowAccessToNodesWithoutACL = false;
+        $plugin->allowUnauthenticatedAccess = false;
+        $fakeServer->addPlugin($plugin);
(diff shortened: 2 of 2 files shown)


mergeMarc:
Travis failed in unrelated unit test (php8): 
```
Sabre\VObject\TimeZoneUtilTest::testTimeZoneBCIdentifiers with data set #131 ('US/Pacific-New')
Exception: DateTimeZone::__construct(): Unknown or bad timezone (US/Pacific-New)
/home/travis/build/sabre-io/dav/vendor/sabre/vobject/tests/VObject/TimeZoneUtilTest.php:174
```

mergeMarc:
I can no longer maintain this PR.

PR #1300, opened 2020-10-23
CalDAV/PropFilter: set empty array as default value for time-range

### LINKED ISSUE #1299: Incorrect default value for time-range CalDAV PropFilter
Currently, the default value for `time-range` in a CalDAV request is set to `false` here: https://github.com/sabre-io/dav/blob/master/lib/CalDAV/Xml/Filter/PropFilter.php

This does not seem to comply with https://tools.ietf.org/html/rfc4791#section-9.9, but it should rather be an empty array.

simonspa:
This changes the default value of `time-range` in CalDAV prop filters from `false` (not matching RFC 4791 https://tools.ietf.org/html/rfc4791#section-9.9) to an empty array.

This closes #1299.

### DIFF
--- a/lib/CalDAV/Xml/Filter/PropFilter.php
+++ b/lib/CalDAV/Xml/Filter/PropFilter.php
@@ -53,7 +53,7 @@ class PropFilter implements XmlDeserializable
             'is-not-defined' => false,
             'param-filters' => [],
             'text-match' => null,
-            'time-range' => false,
+            'time-range' => [],
         ];
 
         $att = $reader->parseAttributes();
--- a/tests/Sabre/CalDAV/Xml/Request/CalendarQueryReportTest.php
+++ b/tests/Sabre/CalDAV/Xml/Request/CalendarQueryReportTest.php
@@ -137,14 +137,14 @@ XML;
                         [
                             'name' => 'UID',
                             'is-not-defined' => false,
-                            'time-range' => false,
+                            'time-range' => [],
                             'text-match' => null,
                             'param-filters' => [],
                         ],
                         [
                             'name' => 'X-PROP',
                             'is-not-defined' => false,
-                            'time-range' => false,
+                            'time-range' => [],
                             'text-match' => null,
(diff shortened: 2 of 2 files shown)


dilyanpalauzov:
See also https://github.com/sabre-io/dav/pull/1275.

phil-davis:
I rebased and fixed the tests.

phil-davis:
@DeepDiver1975 @staabm opinion please on this PR  and PR #1364

lib/CalDAV/Xml/Filter/PropFilter.php
```diff
@@ -53,7 +53,7 @@ public static function xmlDeserialize(Reader $reader)
             'is-not-defined' => false,
             'param-filters' => [],
             'text-match' => null,
-            'time-range' => false,
+            'time-range' => [],
         ];
```
staabm, line 57: I dont have expertise in the part of the codebase - but I think it might be usefull if we could declare constants for this `false`, `null` and `[]` magic values.

simonspa: Might be a good idea, but goes a bit beyond this PR. I am not familiar at all with the SabreDAV code base and I am not willing to stir in it too much by adding constants. :wink:

phil-davis: I can do refactoring for "magic values" after we sort out this PR.

@DeepDiver1975 or @evert are you familiar with this and can give an opinion on changing from `false` to `[]` ?

phil-davis:
> See also #1275.

And #1319

phil-davis:
@DeepDiver1975 any chance to look at this and #1364 please?

simonspa:
Any chance this could get merged? I think it is ready and has been reviewed, the only open suggestion is some larger refactoring that should happen outside the scope of this PR.

DeepDiver1975:
looks good to me :+1:  -> merge

PR #1301, opened 2020-10-23
Add PDO Basic Auth backend

jtsymon:
Equivalent of the existing PDO backend, but with Basic auth. Work around for pimutils/vdirsyncer#749

### DIFF
--- a/examples/sql/mysql.users-basicauth.sql
+++ b/examples/sql/mysql.users-basicauth.sql
@@ -1,0 +1,9 @@
+CREATE TABLE users (
+    id INTEGER UNSIGNED NOT NULL PRIMARY KEY AUTO_INCREMENT,
+    username VARBINARY(50),
+    password_hash VARBINARY(255),
+    UNIQUE(username)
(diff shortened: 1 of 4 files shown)


examples/sql/mysql.users-basicauth.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE users (
+    id INTEGER UNSIGNED NOT NULL PRIMARY KEY AUTO_INCREMENT,
+    username VARBINARY(50),
+    password_hash VARBINARY(255),
+    UNIQUE(username)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+INSERT INTO users (username,password_hash) VALUES
+('admin',  '$2y$10$ovboO1pYLAD0J61zC443X.KfeuE31akctc1PZS4sz7d9Vjzbl1gEi');
```
jtsymon, line 9: This is just the hash of 'admin'

DeepDiver1975:
Any developer can adjust and customize the auth plugins/backends to their needs. The existing database schema and backend implementations are to be seen as reference implementations. 
THX a lot

PR #1305, opened 2020-11-09
CI changes for composer 2.0

phil-davis:
1) fix Travis CI like in https://github.com/sabre-io/skel/pull/17 and the other sabre-io repos
2) postgres on Travis focal (Ubuntu 20.04) is not working. I tried various hints that I found in online discussions, but still no joy. Change the CI for PHP8 to use bionic (Ubuntu 18.04) - that has postgres 12 already pre-installed and ready-to-go.

and this should fetch the new https://github.com/sabre-io/vobject/releases/tag/4.3.3 in CI which will prevent 1 possible unit test error.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,10 +26,8 @@ env:
 matrix:
   include:
     - name: 'PHP8'
-      dist: focal
+      dist: bionic
       php: nightly
-      addons:
-        postgresql: '12'
       env:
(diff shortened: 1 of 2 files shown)


phil-davis:
Travis CI has passed: https://travis-ci.org/github/sabre-io/dav/builds/742359884
The status has not been reported back to GitHub. I will manually set the status for `continuous-integration/travis-ci`

PR #1310, opened 2020-11-20
Support multiget addressdata filter

mstilkerich:
Support filtering returned address object properties in multiget report
    
Analog to addressbook-query report, multiget also supports returning only partial address objects. This can be useful to carddav clients to have fast syncs by excluding large properties like PHOTO and fetching those only when needed.


KristofferFM:
Hi @mstilkerich. Do you know of any reason this patch hasn't been merged yet? I think it is needed, please let me know if there is anything I can help in order to get it into next release.

mstilkerich:
Nope I have no more information than what is in this ticket.

PR #1314, opened 2020-12-12
Explicitly select PHP 8.0 in CI

phil-davis:


phil-davis:
The new php-cs-fixer reports extra stuff. That causes CI errors. See PR #1316

PR #1317, opened 2020-12-28
Update pgsql.calendars.sql

perguth:
Add missing primary key declaration.


staabm:
ohh wow. @phil-davis @DeepDiver1975 is this something you could verify?

phil-davis (approved):
`mysql.calendars.sql` and `sqlite.calendars.sql` both have `id` defined as primary key for table `schedulingobjects`

The change here is the correct way to do this in pgsql, and was missing from this example code.

PR #1321, opened 2021-01-11
Complete checks for non existing/null/false time-range array keys

PhilZ-cwm6:


PhilZ-cwm6:
@phil-davis

After searching in the code, all other occurrences of time-range are properly checked and set to null or false in the code
These are the last two fixes.

I feel this fix is 100% safe to implement as it is again, just syntax fixes to ensure the array key is properly set and not null or false

For code clarity, I went with the `array_key_exists() && $key_value` test because `isset()` is confusing somehow as it accepts boolean == false but not a null value

phil-davis (approved):

Issue #1324, opened 2021-02-04
Trying to request dav endpoint with POST request gives 500 Internal Server Error


phil-davis:
See issue https://github.com/owncloud/core/issues/38378

Sending no Content-Type in a POST can result in a 500 "internal server error" with traceback like:
```
Exception: explode() expects parameter 2 to be string, null given: {
\"Exception\":\"TypeError\",\"Message\":\"explode() expects parameter 2 to be string, null given\",\"Code\":0,\"Trace\":\"
#0 \\/home\\/dipakacharya\\/www\\/core\\/lib\\/composer\\/sabre\\/dav\\/lib\\/DAV\\/Browser\\/Plugin.php(160): explode()
```

If the Content-Type is not specified (null) then the `explode()` explodes!

phil-davis:
Fixed by #1325 
And fixes `owncloud/core` issue https://github.com/owncloud/core/issues/38378

PR #1325, opened 2021-02-04
Correctly process a POST with no Content-Type specified

phil-davis:
Fixes issue #1324

### DIFF
--- a/lib/DAV/Browser/Plugin.php
+++ b/lib/DAV/Browser/Plugin.php
@@ -157,6 +157,9 @@ class Plugin extends DAV\ServerPlugin
     public function httpPOST(RequestInterface $request, ResponseInterface $response)
     {
         $contentType = $request->getHeader('Content-Type');
+        if (!\is_string($contentType)) {
+            return;
+        }
         list($contentType) = explode(';', $contentType);
         if ('application/x-www-form-urlencoded' !== $contentType &&
             'multipart/form-data' !== $contentType) {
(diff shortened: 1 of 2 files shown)


phil-davis:
Note: we should merge #1328 first. Then I will rebase this and get fresh CI.

phil-davis:
@DeepDiver1975 review pelase

lib/DAV/Browser/Plugin.php
```diff
@@ -157,6 +157,9 @@ public function httpGet(RequestInterface $request, ResponseInterface $response)
     public function httpPOST(RequestInterface $request, ResponseInterface $response)
     {
         $contentType = $request->getHeader('Content-Type');
+        if (!\is_string($contentType)) {
+            return;
```
staabm, line 161: Maybe throw a meaningful exception instead?

phil-davis: The code below at line 166 already just returns from this method. That is where things like an empty string for `$contentType` is currently "handled". Having `null` content type or some other thing seems just the same, so I "handled" it the same here. This case just needs to be handled separately before the `explode` because `explode` does not cope with "exploding" `null`

staabm: I see. Thx for elaborating.

seems like this is a bug which would have bern caught with a proper static analysis setup.. we should consider leveling up phpstan and use a baseline

phil-davis:
Note: this was included in PR #1329 which has been merged.

PR #1328, opened 2021-02-04
Test code is not part of vobject distribution - do not run vobject tests in CI

phil-davis:
Issue #1327 

Why do we run the unit tests of the dependencies in the CI here in `sabre/dav` ?

Issue #1327, opened 2021-02-04
CI failing due to no vobject test code


phil-davis:
See demo PR #1326 

https://travis-ci.org/github/sabre-io/dav/jobs/757531864
```
Test directory "tests/../vendor/sabre/vobject/tests/VObject" not found
```

This is because of PR https://github.com/sabre-io/vobject/pull/525 - the test artifacts (and other "rubbish") are not bundled in the release package any more since `sabre/vobject` release 4.3.4 today.

Why do we run the unit tests of dependencies in the CI of `sabre/dav` ?

phil-davis:
Possible "fix" in PR #1328 - "deleted code is debugged code" :)

phil-davis:
Fixed in #1329

PR #1329, opened 2021-02-11
phpstan level 1 and other fixes

phil-davis:
1) Correctly process a POST with no Content-Type specified - code from PR #1325 and issue #1324 
2) Test code is not part of vobject distribution - do not run vobject tests in CI - code from PR #1328 and issue #1327 
3) Increase phpstan to level 1 and add some test coverage to keep codecov happy

### DIFF
--- a/lib/DAV/FSExt/File.php
+++ b/lib/DAV/FSExt/File.php
@@ -41,7 +41,7 @@ class File extends Node implements DAV\PartialUpdate\IPatchSupport
      *
      * The second argument is the type of update we're doing.
      * This is either:
-     * * 1. append
+     * * 1. append (default)
      * * 2. update based on a start byte
      * * 3. update based on an end byte
      *;
@@ -74,6 +74,9 @@ class File extends Node implements DAV\PartialUpdate\IPatchSupport
             case 3:
                 $f = fopen($this->path, 'c');
                 fseek($f, $offset, SEEK_END);
+                break;
+            default:
+                $f = fopen($this->path, 'a');
                 break;
         }
         if (is_string($data)) {
--- a/lib/DAV/Tree.php
+++ b/lib/DAV/Tree.php
@@ -292,6 +292,8 @@ class Tree
             $destinationName = $source->getName();
         }
 
+        $destination = null;
+
         if ($source instanceof IFile) {
             $data = $source->get();
 
--- a/lib/DAV/Xml/Property/Href.php
+++ b/lib/DAV/Xml/Property/Href.php
@@ -39,9 +39,6 @@ class Href implements Element, HtmlOutput
      *
      * You must either pass a string for a single href, or an array of hrefs.
      *
-     * If auto-prefix is set to false, the hrefs will be treated as absolute
-     * and not relative to the servers base uri.
-     *
      * @param string|string[] $hrefs
      */
     public function __construct($hrefs)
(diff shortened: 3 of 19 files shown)


lib/DAV/FSExt/File.php
```diff
@@ -75,6 +75,9 @@ public function patch($data, $rangeType, $offset = null)
                 $f = fopen($this->path, 'c');
                 fseek($f, $offset, SEEK_END);
                 break;
+            default:
```
phil-davis, line 78: phpstan reported that `$f` might not be defined below.
If the caller passes something other than `1` `2` or `3` what should we do?

lib/DAV/Tree.php
```diff
@@ -311,6 +311,8 @@ protected function copyNode(INode $source, ICollection $destinationParent, $dest
             foreach ($source->getChildren() as $child) {
                 $this->copyNode($child, $destination);
             }
+        } else {
+            $destination = null;
```
phil-davis, line 315: phpstan reports that `$destination` might not be defined. So set it here, just in case.

lib/DAV/Xml/Property/Href.php
```diff
@@ -39,9 +39,6 @@ class Href implements Element, HtmlOutput
      *
      * You must either pass a string for a single href, or an array of hrefs.
      *
-     * If auto-prefix is set to false, the hrefs will be treated as absolute
-     * and not relative to the servers base uri.
-     *
```
phil-davis, line 44: The copnstructor had the 2nd parameter removed some time ago. So remove the PHPdoc. And fix the old calls in the test code...

tests/Sabre/CalDAV/Schedule/SchedulingObjectTest.php
```diff
@@ -75,7 +75,6 @@ public function testInvalidArg1()
         $this->expectException('InvalidArgumentException');
         $obj = new SchedulingObject(
             new Backend\MockScheduling([], []),
-            [],
```
phil-davis, line 78: `SchedulingObject` only takes 2 parameters in the constructor. The test code is out-of-date.

tests/Sabre/DAV/DbTestHelperTrait.php
```diff
@@ -58,6 +60,11 @@ public function getDb()
                     }
                     break;
             }
+
+            if (null === $pdo) {
+                $this->markTestSkipped($this->driver.' was not recognised');
+            }
+
             $pdo->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
```
phil-davis, line 68: phpstan reported that `$pdo` might not be set at this point, which is technically possible. Add some code above to deal with that case.

tests/Sabre/DAVACL/PrincipalBackend/Mock.php
```diff
@@ -131,6 +131,8 @@ public function setGroupMemberSet($path, array $members)
     public function updatePrincipal($path, \Sabre\DAV\PropPatch $propPatch)
     {
         $value = null;
+        $principal = false;
+        $principalIndex = null;
```
phil-davis, line 135: phpstan reports that these might not be defined, so initiialize them just in case the `foreach` has an empty loop.

phil-davis:
This includes code from #1325 and #1328 plus increasing the phpstan level from 0 to 1, as suggested in https://github.com/sabre-io/dav/pull/1325#discussion_r571375743

lib/DAV/Tree.php
```diff
@@ -292,6 +292,8 @@ protected function copyNode(INode $source, ICollection $destinationParent, $dest
             $destinationName = $source->getName();
         }
 
+        $destination = null;
```
phil-davis, line 295: phpstan reports that `$destination` might not be defined. So set it here, just in case.

staabm (approved):
Lgtm

PR #1331, opened 2021-02-16
ICS exporter: add startrel parameter to give start relative to now

BastienDurel:
This MR adds a startrel parameter to export plugins, which sets "start" relative to current time, minus interval provided as DateInterval string (eg. "1 year", "18 month", "P2Y" ...)

It also sets componentType to VEVENT if it's not provided, to prevent returning zero elements if start, startrel, end or expand is requested.


DeepDiver1975:
THX a lot. From my POV any developer can add this to an overloaded/extended plugin on theirs own.
I consider this plugin more as a kind of reference implementation and not a full featured exporter.

PR #1332, opened 2021-02-18
feat: Added plugin config option in browser plugin

Aerex:
Context: Allows creating a customization configuration page for plugins. 

The following steps must be done to configure the page. 
  - The plugin must implement the `getConfigBrowser` method. The method should return a raw html page.
  - Update the plugin metadata with `config => true` 
    ```
      function getPluginInfo() {
          return [
            'name'        => $this->getPluginName(),
            'description' => 'The plugin provides synchronization between remote storages and iCal events',
            'link'        => 'https://git.aerex.me/Aerex/baikal-storage-plugin',
            'config'      => true
          ];
        }
    ```
After the following steps are made A cog wheel will appear besides the documentation icon

plugins_config
plugins_page

### DIFF
--- a/lib/DAV/Browser/Plugin.php
+++ b/lib/DAV/Browser/Plugin.php
@@ -144,6 +144,19 @@ class Plugin extends DAV\ServerPlugin
                 $response->setBody(
                     $this->generatePluginListing()
                 );
+            case 'pluginConfigs':
+                if (!isset($getVars['pluginName'])){ 
+                  $response->setStatus(400);
+                  $response->setBody('Missing pluginName query param');
+                  return false;
+                }
+                $response->setStatus(200);
+                $response->setHeader('Content-Type', 'text/html; charset=utf-8');
+
+                $response->setBody(
+                    $this->generatePluginConfigs($getVars['pluginName'])
+                );
+
 
                 return false;
         }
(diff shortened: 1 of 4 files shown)


DeepDiver1975:
Browser support is a pure developer feature. From my pov devs should know on how to configure various plugins in the code base. I feel like this is not adding much benefit but adds even more maintenance burden on us.
THX a lot

Aerex:
@DeepDiver1975 I understand your reasoning but the purpose of this plugin to give the users the ability to configure the plugin who **may not**  have the technical skills to do so. 

I have created a plugin to allow syncing calendar events with other storages. 

Could you reconsider or what can I can to improve on this feature? Also, could elaborate on how this change cane be a burden to maintain?

DeepDiver1975:
> Also, could elaborate on how this change cane be a burden to maintain?

The team behind sabre is very limited on time to keep the library up and running. Any additional code added is increasing the effort we need to put in. We simply cannot accept new features.



> to give the users the ability to configure the plugin who **may not** have the technical skills to do so.

sabre libs are for developers - not for end users. Please implement any user facing features within your dedicated project.

Aerex:
I see. Understood. Thanks for your response.

PR #1335, opened 2021-03-12
Fix: don't remove lock on dir when deleting a child node

tbsmark86:
Current behavior:
1) lock /somedir/
2) delete /somedir/somefile.txt (with token)
3) put /somedir/newfile.txt (with token)
   => ERROR precondition failed

Expected/Fixed behavior:
Step 3) should be executed under the LOCK.


phil-davis:
Assigned myself so that I find this easily to followup.

phil-davis (approved):
LGTM.
Note: ownCloud is using its own implementation of locking and so I could not reproduce this on ownCloud - I don't think it ever executes this code.

Issue #482, opened 2014-07-21
Cannot add custom properties to a propFind for allprops


ghost:
Encountered when converting the [SambaDAV](https://github.com/bokxing-it/sambadav) plugin to use the SabreDAV-2.0.4 `propFind` and `propPatch` events. SabreDAV has a fixed [list of properties](https://github.com/fruux/sabre-dav/blob/380d6cea83e657a5fd5a89bed126418ec8742f78/lib/DAV/PropFind.php#L51) that it will return for an "allprops" request. Fair enough. But if we add a propFind plugin that calls 'set' for a nonstandard property x, I expect property x to be hard added to the response. Instead, SabreDAV decides that property x is not in the pre-approved list, and silently drops it.

Background: in [SambaDAV](https://github.com/bokxing-it/sambadav) we always add the nonstandard `{DAV:}ishidden`, `{DAV:}isreadonly` and Win32 flag properties during a propfind, to support mounting as a Windows Network Share. I expected this (demo) code to set the property:

``` php
class MSPropertiesPlugin extends DAV\ServerPlugin
{
    public function initialize (DAV\Server $server)
    {
        $server->on('propFind', [ $this, 'propFind' ]);
    }
    public function propFind (DAV\PropFind $propFind, DAV\INode $node)
    {
        // Hide all the things!
        $propFind->set('{DAV:}ishidden', '1');
    }
}
```

Instead the property is ignored (unless we add it to SabreDAV's "allprops" list by hand).

PR #1344, opened 2021-05-14
Fix propfind type of subnodes with depth >= 1

n-peugnet:
With properties manually added in a plugin using the `propFind` event like in this example: https://github.com/sabre-io/dav/issues/482 and with depth >= 1, the subnodes' propfind's type was incorrectly deduced.
This resulted in custom properties only being added to the first node and not on the subnodes.

### DIFF
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -895,7 +895,7 @@ class Server implements LoggerAwareInterface, EmitterInterface
         }
 
         $propertyNames = $propFind->getRequestedProperties();
-        $propFindType = !empty($propertyNames) ? PropFind::NORMAL : PropFind::ALLPROPS;
+        $propFindType = !$propFind->isAllProps() ? PropFind::NORMAL : PropFind::ALLPROPS;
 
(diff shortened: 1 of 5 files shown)


n-peugnet:
I added a test case and we can see that it fails on master but not with my patch :

```
There was 1 failure:

1) Sabre\DAV\ServerPropsTest::testPropFindEmptyBodyDepth1Custom
Response should contain 5 elements
Failed asserting that 1 matches expected 5.

/home/nicolas/Source/php/dav/tests/Sabre/DAV/ServerPropsTest.php:93
```

kesselb:
Hey @n-peugnet :wave:  

please rebase your branch to have a recent CI run.

n-peugnet:
Done @kesselb

n-peugnet:
@phil-davis, could you maybe approve this PR to run the GitHub workflow ?

I have run the tests locally and everything is working fine.

phil-davis (approved):
While I  was looking, I refactored some identical preg_replace "magic" in the test code so that it is in a separate function.

PR #1360, opened 2021-11-02
Trigger CI

phil-davis:
to make the new GitHub Actions run


phil-davis:
The latest `phpstan` reports problems. I will make a PR to fix those.

PR #1362, opened 2021-11-02
Fix return types

### LINKED ISSUE #1361: Latest phpstan is reporting missing return statements
phpstan is reporting various functions that have return types documented but do not always return the documented type.

This is causing CI to fail.

phil-davis:
Fixes #1361 

I have tried to minimize changes to real code, and o mostly just delete or adjust PHPdoc to match reality. That gets CI green with the minimum impact.

### DIFF
--- a/lib/CalDAV/Plugin.php
+++ b/lib/CalDAV/Plugin.php
@@ -243,7 +243,7 @@ class Plugin extends DAV\ServerPlugin
      * @param mixed  $report
      * @param mixed  $path
      *
-     * @return bool
+     * @return bool|null
      */
     public function report($reportName, $report, $path)
     {
--- a/lib/CalDAV/Schedule/Plugin.php
+++ b/lib/CalDAV/Schedule/Plugin.php
@@ -735,9 +735,7 @@ class Plugin extends ServerPlugin
 
     /**
      * This method is responsible for parsing a free-busy query request and
-     * returning it's result.
-     *
-     * @return string
+     * returning its result in $response.
      */
     protected function handleFreeBusyRequest(IOutbox $outbox, VObject\Component $vObject, RequestInterface $request, ResponseInterface $response)
     {
--- a/lib/CalDAV/Subscriptions/Subscription.php
+++ b/lib/CalDAV/Subscriptions/Subscription.php
@@ -75,7 +75,7 @@ class Subscription extends Collection implements ISubscription, IACL
     /**
      * Returns the last modification time.
      *
-     * @return int
+     * @return int|null
      */
     public function getLastModified()
     {
--- a/tests/Sabre/CalDAV/Backend/AbstractTest.php
+++ b/tests/Sabre/CalDAV/Backend/AbstractTest.php
@@ -140,6 +140,8 @@ class AbstractMock extends AbstractBackend
                     ],
                 ];
         }
+
+        return [];
     }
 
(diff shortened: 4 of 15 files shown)


phil-davis:
```
------ -------------------------------------------------------- 
  Line   tests/Sabre/CalDAV/ExpandEventsDTSTARTandDTENDTest.php  
 ------ -------------------------------------------------------- 
  100    Variable $vObject might not be defined.                 
 ------ -------------------------------------------------------- 
```

phpstan fails in CI, but I don't get this fail locally.

ChristophWurst (approved):

lib/CalDAV/Plugin.php
```diff
@@ -264,6 +264,8 @@ public function report($reportName, $report, $path)
 
                 return false;
         }
+
+        return true;
```
phil-davis, line 268: what to do about these sort of things?
I guess that this code is never supposed to get past the switch cases. But what if it does? return true or false?

lib/CalDAV/Schedule/Plugin.php
```diff
@@ -735,9 +735,7 @@ public function outboxRequest(IOutbox $outboxNode, RequestInterface $request, Re
 
     /**
      * This method is responsible for parsing a free-busy query request and
-     * returning it's result.
-     *
-     * @return string
```
phil-davis, line 740: This function has no return statements at all. Why is the PHPdoc saying that it should return a string?

lib/CalDAV/Subscriptions/Subscription.php
```diff
@@ -75,13 +75,15 @@ public function getName()
     /**
      * Returns the last modification time.
      *
-     * @return int
+     * @return int|false
```
staabm, line 78: Before this change the method returned nothing, which means null.

lib/CalDAV/Plugin.php
staabm: Maybe throw a exception?

lib/CalDAV/Subscriptions/Subscription.php
phil-davis, line 78: Callers seem to expect this return a falsy value if `lastmodified` is not set.

phil-davis: I will replace false with null and revert the addition of `return false;` - that will be super-safe to do.

tests/Sabre/CalDAV/Backend/AbstractTest.php
```diff
@@ -140,6 +140,8 @@ public function getCalendarObjects($calendarId)
                     ],
                 ];
         }
+
+        return [];
```
phil-davis, line 144: From here, the changes are test code. 
We can put this sort of `return` at the end of a function to keep phpstan happy. It will never be executed, but that is no problem.

lib/CalDAV/Plugin.php
phil-davis: I don't really know the calling path of all these plugins. If `$reportName` does not match any of the switch cases, then maybe the function is supposed to return "nothing" and not explode?

Is the existing omission of a return at the end of the function a bug or a feature?

phil-davis: I got rid of this and updated the PHPdoc to allow null.

PR #1363, opened 2021-11-14
Sabre/Xml tests are no longer published

phil-davis:
https://github.com/sabre-io/xml/pull/195 added `.gitattributes` that specified:
```
/tests export-ignore
```

So `vendor/sabre/xml` no longer contains the `tests` folder.

Remove it from `phpunit.xml` - we can't run those tests here in `dav`

CI is currently red in `master` because of this.

### DIFF
--- a/tests/phpunit.xml
+++ b/tests/phpunit.xml
@@ -10,9 +10,6 @@
 >
 
   <testsuites>
-      <testsuite name="sabre-xml">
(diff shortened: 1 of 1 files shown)


phil-davis:
@staabm @evert @DeepDiver1975 is there a requirement to run the unit tests from other repos in CI here?

PR #1364, opened 2021-11-16
More CalDAV empty-array time-range changes

phil-davis:
This is on top of #1300 

There are other places where `time-range` is set to false or null by default in various places.

Should these all be an empty array?


phil-davis:
@DeepDiver1975 do you have knowledge/opinion about this?
And #1300

PR #1365, opened 2021-11-17
add params for put interface

phil-davis:
This is a rebase of #1290 

That PR comes from the `master` branch in https://github.com/yrong/dav and I don't want to mess with that (and maybe I can't)

### DIFF
--- a/lib/DAV/IFile.php
+++ b/lib/DAV/IFile.php
@@ -38,7 +38,7 @@ interface IFile extends INode
      *
      * @return string|null
      */
-    public function put($data);
+    public function put($data,$params=null);
 
     /**
      * Returns the data.
--- a/lib/DAV/IFile.php
+++ b/lib/DAV/IFile.php
@@ -35,6 +35,7 @@ interface IFile extends INode
      * return an ETag, and just return null.
      *
      * @param resource|string $data
+     * @param object|null     $params
      *
      * @return string|null
      */
--- a/lib/DAV/IFile.php
+++ b/lib/DAV/IFile.php
@@ -38,7 +38,7 @@ interface IFile extends INode
      *
      * @return string|null
      */
-    public function put($data,$params=null);
+    public function put($data, $params = null);
 
     /**
      * Returns the data.
--- a/lib/CalDAV/CalendarObject.php
+++ b/lib/CalDAV/CalendarObject.php
@@ -95,7 +95,7 @@ class CalendarObject extends \Sabre\DAV\File implements ICalendarObject, \Sabre\
      *
      * @return string
      */
-    public function put($calendarData,$params=null)
+    public function put($calendarData, $params = null)
     {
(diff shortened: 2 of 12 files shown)


phil-davis:
I added PHPdoc and some minor changes in an extra commit.

phil-davis:
@yrong do you still need this?
Please advise the details...

lib/DAV/IFile.php
```diff
@@ -35,10 +35,11 @@ interface IFile extends INode
      * return an ETag, and just return null.
      *
      * @param resource|string $data
+     * @param object|null     $params
      *
      * @return string|null
      */
-    public function put($data);
+    public function put($data, $params = null);
```
staabm, line 42: this param must be optional, to prevent a BC break. 👍

staabm (approved):
lgtm

lib/DAV/IFile.php
dpfaffenbauer: this still is a bc break: `[php] Fatal Compile Error: Declaration of Pimcore\Model\Asset\WebDAV\File::put($data) must be compatible with Sabre\DAV\File::put($data, $params = null) ["exception" => Symfony\Component\ErrorHandler\Error\FatalError { …}]`

phil-davis: That's annoying! 
It can nearly work because `put($data)` can be happily written when calling an instance of the parent and/or the child class. If PHP 7.2+ didn't check this, then actually this code could still work.

But if someone writes new code that calls an instance a `IFile` then they should be able to write `put($data, $params)`  and pass the new parameter, if they wish. But then at run-time the object might be an instance of the child class, and that object will not know anything about the 2nd parameter. But the child should function as a fully-working drop-in replacement for the parent class - it doesn't.

Therefore this change is dangerous for future code (but could have worked for everyone's existing code, if PHP "let it past").

A good explanation of this stuff is in https://platform.sh/blog/watch-your-inheritance/

phil-davis: Reverted in PR #1375

phil-davis: Revert is released in https://github.com/sabre-io/dav/releases/tag/4.2.3

@dpfaffenbauer thanks for reporting! I will ask if that extra parameter to `put($data)` is really a needed thing. If it is, then we need to think if there is any BC way to implement it, or if it would have to be part of a major version.

dpfaffenbauer: Thanks a lot for reacting so fast :) I owe you a beer now. 

Regarding BC, add the new param like this /* $params = null */ and use func_get_args to check if there is a second param

phil-davis: > Thanks a lot for reacting so fast :) I owe you a beer now.

Actually, I owe you a beer for reporting this quickly, so that it could be reverted  before too many people bump their dependencies.

Issue #1370, opened 2021-11-30
Inconsistent return type for \Sabre\DAVACL\PrincipalBackend\BackendInterface::findByUri


come-nc:
Hello,

See the docblock for this function:
```
    /**
     * Finds a principal by its URI.
     *
     * This method may receive any type of uri, but mailto: addresses will be
     * the most common.
     *
     * Implementation of this API is optional. It is currently used by the
     * CalDAV system to find principals based on their email addresses. If this
     * API is not implemented, some features may not work correctly.
     *
     * This method must return a relative principal path, or null, if the
     * principal was not found or you refuse to find it.
     *
     * @param string $uri
     * @param string $principalPrefix
     *
     * @return string
     */
```

The comment says that the function may return null, but then there is `@return string` instead of `@return string|null`.
We implement it as returning `string|null` and this triggers psalm errors in our CI.

DeepDiver1975:
As far as I can tell from implementing this interface as well: you should return and empty string instead of null.

DeepDiver1975:
> * This method must return a relative principal path, or null, if the
>      * principal was not found or you refuse to find it.

doc wrong or annotation wrong? :see_no_evil:

DeepDiver1975:
ah - already taken care of: https://github.com/sabre-io/dav/pull/1362

come-nc:
> ah - already taken care of: #1362

In the abstract class, but not in the interface: https://github.com/sabre-io/dav/blob/master/lib/DAVACL/PrincipalBackend/BackendInterface.php#L115

DeepDiver1975:
Feel free to submit a PR to address this - THX a lot

PR #1371, opened 2021-11-30
Fix phpdoc return type of findByUri in BackendInterface

come-nc:
Fix phpdoc return type of findByUri

Fixes https://github.com/sabre-io/dav/issues/1370


phil-davis (approved):

phil-davis:
I published https://github.com/sabre-io/dav/releases/tag/4.2.1 for this just now. It might as well be "out there" so people don't get warnings from code analysers.

PR #1373, opened 2021-12-08
Allow psr/log ^1.0 || ^2.0 || ^3.0

fezfez:


phil-davis:
@fezfez this is probably "a good thing" - composer can decide which major version of `psr/log` it should use.
Please write a few words in the PR describing how this is "a good thing" (for example, maybe it helps when using `sabre/dav` along with some mix of other dependencies?) Then it will be easy for future readers to see why the change was made.

fezfez:
If you have locked psr/log:^2.0, the sabre/dev:^3.0 will be installed, witch is not working (in my case) in php ^8.0

phil-davis:
https://github.com/sabre-io/dav/releases/tag/4.2.2 is available

PR #1375, opened 2021-12-09
Revert PR 1365 "add params for put interface"

phil-davis:
See comment https://github.com/sabre-io/dav/pull/1365#discussion_r765559132

That change was not strictly backward-compatible. Revert it for now, and I will release a new patch version.

Then we can discuss if the change is really wanted, and if there is a backward-compatible way to do it.


phil-davis:
Note: I also noticed that the previous code broke stuff in `owncloud/core` 
`PHP Fatal error:  Declaration of OCA\DAV\Connector\Sabre\File::put($data) must be compatible with Sabre\DAV\IFile::put($data, $params = NULL) in /drone/src/apps/dav/lib/Connector/Sabre/File.php on line 138`

Issue #1351, opened 2021-09-12
Browser plugin : HTTP2 error with SSL (solution given)


cyprienhennon:
When using the browser plugin with SSL and HTTP2, the assets (css, icons, ...) are not loaded and the page loading is very slow.
I find the solution : In the plugin.php file in the browser directory, modify the function serveAsset() by removing the line $this->server->httpResponse->setHeader('Content-Length', filesize($assetPath));
Without this header, HTTP2 works.
And everything is working all right !

phil-davis:
Unit tests fail because they test for the Content-Length header.
@cyprienhennon please adjust the unit tests.

@staabm @DeepDiver1975 please give your thoughts. Is this actually a problem with HTTP2 SSL... implementation, rather than with `sabre/dav` ? Why would we stop sending Content-Length?

ZatsuneNoMokou:
If you have this weird problem (for exemple a apache/php cgi like setup), you can add this on your side before the `$server->start()` :

```php
$server->on('afterMethod:get', function (Request $request, Response $response) {
    $response->removeHeader('Content-Length');
});
```

This will remove the content-length without having to change the library code (until there's a better option).

PR #1379, opened 2021-12-17
Update browser plugin

cyprienhennon:
Deleting the line 585 $this->server->httpResponse->setHeader('Content-Length', filesize($assetPath)); to add compatibility with HTTP2 and SSL

With this line, the server works good, but in the browser, if the user uses SSL ans the server HTTP2, the page loading is very slow and the browser throws an error ERR_HTTP2_PROTOCOL_ERROR.

### DIFF
--- a/lib/DAV/Browser/Plugin.php
+++ b/lib/DAV/Browser/Plugin.php
@@ -582,7 +582,6 @@ HTML;
         }
 
         $this->server->httpResponse->setHeader('Content-Type', $mime);
-        $this->server->httpResponse->setHeader('Content-Length', filesize($assetPath));
         $this->server->httpResponse->setHeader('Cache-Control', 'public, max-age=1209600');
         $this->server->httpResponse->setStatus(200);
         $this->server->httpResponse->setBody(fopen($assetPath, 'r'));


cyprienhennon:
Issue #1351 Browser plugin : HTTP2 error with SSL (solution given)

phil-davis:
Unit tests fail because they test for the Content-Length header.
@cyprienhennon please adjust the unit tests.

@staabm @DeepDiver1975 please give your thoughts. Is this actually a problem with HTTP2 SSL... implementation, rather than with `sabre/dav` ? Why would we stop sending Content-Length?

staabm:
it could be a problem, when php is configured to ouput a compressed stream (php-ini: zlib.output_compression) but our content-length is not calculated according to zlib but in raw bytes.

not sure this is the source of the initial reported problem though.

```
        if (!ini_get('zlib.output_compression')) {
            header('Content-Length: ' . filesize($assetPath));
        }
```

in case the browser sends chunked-encoded requests our current length implementation would also be wrong (since we are not advertising chunked length encoding support via http this should not be the case though - but maybe the reporter is overrwriting http headers on the webserver level or at a proxy level)

cyprienhennon:
I just tested the @staabm code :
`        if (!ini_get('zlib.output_compression')) {
            header('Content-Length: ' . filesize($assetPath));
        }`
and it doesn't solve the problem.
I'm not an expert but i think it's because HTTP/2 "pushes" the assets with the main request.
In HTTP/1.1, the server send an answer to the client request, and the client identify some assets it needs to show the page, an make a new request for each asset to have it.
In HTTP/2 the server identify some assets the clients gonna need, and sends the answer + some assets in the "same" time, without the client had to make a request for theese assets.
By indicating the content length of an asset, we give a length that is not correct, because the server gonna send multiple data.
I think we should let the server manage this header depending on the protocol because in HTTP/1.1, the server will calculate the content length and send the header by itself, and in HTTP/2 the server will manage the assets sending.
It's only the question of the browser plugin's assets, and not about the other files send by sabre.
_I'm not sure that my english is perfect, I hope you will understand me ... 😅_

staabm:
I am not yet sure what the concrete problem is, but I am pretty sure the problem we see is not related to http2 server push. 

IIRC sabre* libs do not send the push headers and most browsers removed support for http2 push these days

cyprienhennon:
OK, as I said i'm not an expert ..
I only noticed that with this line, Chrome is not charging any asset (CSS, icons, logo) and throws the net::ERR_HTTP2_PROTOCOL_ERROR, and without the page rendering is complete.
So I extends the browser plugin class in my app to solve the problem, and I propose this modification.
But sabre works well without this change, personally, I just use the browser plugin for the dev tests, I don't know if many other users have the same problem.

staabm:
got a different idea what the cause of the problem could be.
could you check whether you have a file in your codebase which contains leading spaces before php-open tags or trailiing whitespaces after php-closing tags?

these spaces could affect the content beeing rendered to the browser and lead to a mismatch of content-length vs. actual content

cyprienhennon:
@staabm Thanks for your answer
I checked my php files (and i'm just working with php on the server), and none had a space before the php open-tag, and I never use php close-tags
I tried also to upload a file via the browser plugin, and the size is equal, and the sha-1 hash before and after is equal too ...

staabm:
I see. Seems a bit strange but I am fine with this PR as is. 
 
@phil-davis any opinion?

phil-davis:
@DeepDiver1975 what do you think?

PR #1381, opened 2022-01-10
[tests-only] Break tests to check CI

phil-davis:
I want to see which PDO tests are really running against which database.

Related to Baikal issue https://github.com/sabre-io/Baikal/issues/1082 where I hoped that the problem would already have been caught by CI here.


phil-davis:
Good - the `Sabre\CalDAV\Backend` PDO unit tests are running.

PR #1391, opened 2022-02-21
add pear mail usage for smtp mailing

epsilon-0:
Signed-off-by: Aisha Tammy <floss@bsd.ac>


DeepDiver1975:
I honestly see no need in this - if you want to use a different mail implementation you can sub class the plugin and implement mail delivery as you like.

PR #1396, opened 2022-03-29
Include "before" and "after" copy events as with move events

jvillafanez:
The same way "beforeMove" and "afterMove" events are triggered, the new "beforeCopy" and "afterCopy" events will be triggered during a COPY request.


DeepDiver1975:
There should be unit tests for event handling. Please have a look and extend the tests. Thx 🙏

jvillafanez:
I don't think I can rise the project coverage taking into account that the patch has 100% coverage. There might be tests missing outside of the scope of this PR.

mmattel:
@DeepDiver1975 merging?

PR #1404, opened 2022-06-14
Fix encoding detection on PHP 8.1

come-nc:
Also get rid of deprecated utf8_encode

Spin-off of https://github.com/sabre-io/http/pull/182


mbiebl:
Is a new sabre/dav release planned soonish with this fix? As mentioned in #1405, I've run into this issue with vCards containing names with Slovenian characters.

phil-davis (approved):

PR #1408, opened 2022-07-05
add LDAP backend + cyrus sasl substitution

epsilon-0:
Also add stub functionality for auto user creation.

This is useful when adding a user in LDAP and it can get automatically added in the Baikal database.


epsilon-0:
I thought that as the IMAP auth plugin is present here, the LDAP could also be added, instead of everyone needing to implement one themselves. 

Plus the cyru-sasl style substitution function is also commonly used with LDAP backends, which I feel is also something that could be added here for general use, rather than everyone having to re-implement the same and increasing churn.

PR #1421, opened 2022-10-09
PHP 8.2 in CI

phil-davis:
- add PHP 8.2 to CI
- also run the standard CI (non-streaming) with PHP 8.1 (which already runs with the streaming option)


phil-davis:
That was easy - everything still passes with PHP 8.2

PR #1425, opened 2022-10-22
Fix types in Service.php (type error in php 8.1).

TCKnet:


lib/DAV/Xml/Service.php
```diff
@@ -40,7 +40,7 @@ class Service extends \Sabre\Xml\Service
      *
      * @var array
```
phil-davis, line 41: php-cs-fixer wants this PHP-doc to be removed (because now the array type is fully declared below)

phil-davis:
This does not work with older PHP versions - see unit test results. We have not (yet) dropped those old versions, but are planning to do so "real soon now" (tm).

@TCKnet what exactly happens in PHP 8.1 that needs the type to be declared here?

TCKnet:
On the minimal example with php 8.1.11:
<?php
error_reporting(0);
class Service { public array $map = []; }
class DavService extends Service { public $map = ['k' => 'v']; }
?>
$ php test.php
Fatal error: Type of DavService::$map must be array (as in class Service) in test.php on line 4

A workaround could be to remove the type in  Service: this doesn't generate an issue 
class Service { public $map = []; }

This however doesn't seem documented, there is an issue about function returns, but not with properties :-/
https://www.php.net/manual/en/migration81.incompatible.php

However all the rest seems to parse fine with PHP 8.1.

DeepDiver1975:
Please open an issue with detailed reproduction steps. Closing this PR for now. THX a lot

PR #1432, opened 2022-12-12
Remove unused itipMessage property

come-nc:
Unless I missed something this property is never set or get?
Spotted by psalm.


ChristophWurst (approved):
PHPStorm finds no usage either

PR #1439, opened 2023-01-20
Allow recent sabre major version dependencies

phil-davis:


phil-davis:
Test directory "tests/../vendor/sabre/http/tests/HTTP" not found

The unit tests are trying to also run the sabre/http unit tests. But I think that those are no longer distributed in the release. Needs investigation...

phil-davis:
Fatal error: Type of Sabre\DAV\Xml\Service::$elementMap must be array (as in class Sabre\Xml\Service) in /home/runner/work/dav/dav/lib/DAV/Xml/Service.php on line 14
https://github.com/sabre-io/dav/actions/runs/3966326852/jobs/6797002225

It wants us to write:
```
public array $elementMap
```
o maybe 
But declaring the type of a property is a PHP 7.4 feature. So we can't do that until we drop PHP 7.1 7.2 7.3.

And I am waiting to only do that after getting sabre/vobject code updates done. So maybe we can't use the newer sabre/xml just yet.

phil-davis:
Same problem with `class ResponseMock extends Response` 
We need to say `public int $status;` but that is a PHP 7.4 thing.

I think that I will stop trying to do any of this right now.

PR #1441, opened 2023-01-20
Bump php-cs-fixer to v3

phil-davis:


phil-davis:
The new php-cs-fixer v3 needs at least PHP 7.4 - that would be OK because we can fixup CI to only run php-cs-fixer on PHP 7.4 and up.
But I think that there will be similar restrictions on updating phpstan, and it would be nice to run that on all the PHP versions.
Now I remember why we were a bit stuck here. This is easiest if it waits for us gto drop PHP 7.1 7.2 7.3 support.

PR #1442, opened 2023-01-23
Upgrading sabre/vobject 4.5.3

phil-davis:
Just checking if CI passes with the new vobject 4.5.3 release.


phil-davis:
Good - vobject 4.5.3 works with current dav

Issue #1445, opened 2023-01-27
PHP 8.2 dynamic properties deprecation warnings


AaronVanGeffen:
In PHP 8.2, dynamic class properties [have been deprecated](https://www.php.net/releases/8.2/en.php#deprecate_dynamic_properties). Apparently, Sabre\DAV still uses some in at least two locations, leading to errors below.

Baikal is working again after quickly patching the two classes. I do not currently have time to upstream my changes, but I thought I'd report these findings either way.

```
PHP message: ErrorException: Creation of dynamic property Sabre\DAV\Xml\Property\Href::$autoPrefix is deprecated in /usr/share/webapps/baikal/vendor/sabre/dav/lib/DAV/PropertyStorage/Backend/PDO.php:100
Stack trace:
#0 [internal function]: Baikal\Framework::exception_error_handler()
#1 /usr/share/webapps/baikal/vendor/sabre/dav/lib/DAV/PropertyStorage/Backend/PDO.php(100): unserialize()
#2 /usr/share/webapps/baikal/vendor/sabre/dav/lib/DAV/PropertyStorage/Plugin.php(84): Sabre\DAV\PropertyStorage\Backend\PDO->propFind()
#3 /usr/share/webapps/baikal/vendor/sabre/event/lib/WildcardEmitterTrait.php(89): Sabre\DAV\PropertyStorage\Plugin->propFind()
#4 /usr/share/webapps/baikal/vendor/sabre/dav/lib/DAV/Server.php(1052): Sabre\DAV\Server->emit()
#5 /usr/share/webapps/baikal/vendor/sabre/dav/lib/DAV/Server.php(984): Sabre\DAV\Server->getPropertiesByNode()
#6 /usr/share/webapps/baikal/vendor/sabre/dav/lib/DAV/Server.php(1662): Sabre\DAV\Server->getPropertiesIteratorForPath()
#7 /usr/share/webapps/baikal/v
```

```
PHP message: ErrorException: Creation of dynamic property Sabre\CardDAV\Xml\Request\AddressBookMultiGetReport::$addressDataProperties is deprecated in /usr/share/webapps/baikal/vendor/sabre/dav/lib/CardDAV/Xml/Request/AddressBookMultiGetReport.php:104
Stack trace:
#0 /usr/share/webapps/baikal/vendor/sabre/dav/lib/CardDAV/Xml/Request/AddressBookMultiGetReport.php(104): Baikal\Framework::exception_error_handler()
[...]

PR #1447, opened 2023-01-31
convertDeprecationsToExceptions during PHPunit tests

phil-davis:
https://phpunit.readthedocs.io/en/9.5/configuration.html#the-convertdeprecationstoexceptions-attribute


phil-davis:
I was hoping that this would fail on PHP 8.2 because of the problems reported in issue #1445 

But it passed - needs investigation.

phil-davis:
> But it passed - needs investigation.

The deprecation things in recent PHP 8.* have been fixed in other PRs. But, IMO, we might as well have this extra setting in the unit tests. It may help us notice a deprecation in future.

Issue #1458, opened 2023-03-26
using root directory C:\ fails on Windows 10


divinity76:
i have a fairly standard
```php
<?php
declare (strict_types = 1);
use Sabre\DAV;

// The autoloader
require 'vendor/autoload.php';

// Now we're creating a whole bunch of objects
$rootDirectory = new DAV\FS\Directory('C:');

// The server object is responsible for making sense out of the WebDAV protocol
$server = new DAV\Server($rootDirectory);
require_once __DIR__ . "/authentication.php";
// If your server is not on your webroot, make sure the following line has the
// correct information
$server->setBaseUri('/webdav/');

// The lock manager is reponsible for making sure users don't overwrite
// each others changes.
$lockBackend = new DAV\Locks\Backend\File('data/locks');
$lockPlugin = new DAV\Locks\Plugin($lockBackend);
$server->addPlugin($lockPlugin);

// This ensures that we get a pretty index in the browser, but it is
// optional.
$server->addPlugin(new DAV\Browser\Plugin());

// All we need to do now, is to fire up the server
$server->exec();
```
and have tried several variations of
```
$rootDirectory = new DAV\FS\Directory('C:');
$rootDirectory = new DAV\FS\Directory('C:\\');
$rootDirectory = new DAV\FS\Directory('C:/');
```
and pretty much all of them fail with some variation of
```xml
<?xml version="1.0" encoding="utf-8"?>
<d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
  <s:sabredav-version>3.2.2</s:sabredav-version>
  <s:exception>Sabre\DAV\Exception\NotFound</s:exception>
  <s:message>File with name C:/DumpStack.log.tmp could not be located</s:message>
</d:error>
```

DeepDiver1975:
solution given in #1460

PR #1460, opened 2023-03-26
hack to allow mounting C:\ on Windows 10

divinity76:
for some reason these files exist but fails file_exists("C:\\hiberfile.sys"); which then causes an exception in getChild(), i think it's a bug in PHP's file_exists(), where file_exists() return false if a file "exists but is not readable", and IIRC someone raised the issue on the php bugtracker, but the php core devs refused to fix it for "backwards compatibility reasons" ... don't remember the bug report number though.

anyway, this hack allows mounting C:\ on windows 10. Related issue: https://github.com/sabre-io/dav/issues/1458

### DIFF
--- a/lib/DAV/FS/Directory.php
+++ b/lib/DAV/FS/Directory.php
@@ -100,17 +100,17 @@ class Directory extends Node implements DAV\ICollection, DAV\IQuota {
             \FilesystemIterator::CURRENT_AS_SELF
           | \FilesystemIterator::SKIP_DOTS
         );
-        $isWindows = (PHP_OS === 'WINNT');
+        $enableCMountHack = (PHP_OS === 'WINNT' && $this->path === 'C:');
         foreach ($iterator as $entry) {
-            if ($isWindows) {
+            if ($enableCMountHack) {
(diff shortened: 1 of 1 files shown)


divinity76:
related PHP bugtracker issue: https://bugs.php.net/bug.php?id=78285
quoting [cmb@php.net](https://people.php.net/cmb):
>Actually, it is to be expected that file_exists() is much slower
than is_dir() (when the realpath cache is enabled), because the
former checks whether the file may be accessed, and since this
information is not cached, it is done for each and every call.
>
>Changing to doc problem, since this behavior is not described in
the manual.

DeepDiver1975:
I honestly do not consider mounting `C:/` a reasonable use case which shall be handled in the core feature set of sabre/dav.

Who knows what other files will one day exist in `C:/`?
What if someone wants to mount `D:/` which was the same files?

I suggest you look into the plugin architecture of sabre/dav and implement a plugin for your explicit case.
As idea you can look into  https://github.com/sabre-io/dav/blob/master/lib/DAV/TemporaryFileFilterPlugin.php.

THX

PR #1475, opened 2023-06-26
Bump composer dependencies

phil-davis:
just checking that CI passes (sabre/http 5.1.7 was released just now)

We don't actually need these changes.


phil-davis:
Good - the unit tests pass with `sabre/http` 5.1.7 and all the PHP versions 7.1 to 8.2

PR #1486, opened 2023-09-02
Checking for destory() function

toteph42:
Would be great, if you can accept. My implementation of Sabre broke due to missing destroy functions.


DeepDiver1975:
Which class misses the destroy() function?

DeepDiver1975 (changes requested):
- answer question of reviewers
- don't add IDE specific files
- split changes to individual pull requests
- don't change the project name in composer

Issue #1487, opened 2023-09-18
Response must not contain `stats` **and** `propstat`


susnux:
As per [rfc2518](http://www.webdav.org/specs/rfc2518.html#ELEMENT_response) the *response* element
* **must** contain a `href`
* *might* contain a `responsedescription`
* contain **either** `(status|href*)` **OR** `propstat+`

So this code might produce invalid responses:
https://github.com/sabre-io/dav/blob/85b33f7c4b597bdb2a90e6886a48c5723e767062/lib/DAV/Xml/Element/Response.php#L115

PR #1488, opened 2023-09-18
fix: The WebDAV response element must only contain `propstat` OR `status` element(s)

susnux:
* Resolves https://github.com/sabre-io/dav/issues/1487

The response element should look like, see [section-14.24](https://datatracker.ietf.org/doc/html/rfc4918#section-14.24) :
`<!ELEMENT response (href, ((href*, status)|(propstat+)), error?, responsedescription? , location?) >`

Meaning it
* Must contain a `href` :heavy_check_mark: 
* Might contain `error`, `responsedescription` and `location` (not implemented by us) :white_check_mark: 
* Must **EITHER** contain a status **OR** one or more `propstat` elements

So we have to ensure the status is only set if there is no `propstat`.

### DIFF
--- a/lib/DAV/Xml/Element/Response.php
+++ b/lib/DAV/Xml/Element/Response.php
@@ -40,7 +40,7 @@ class Response implements Element
      *
      * This is currently only used in WebDAV-Sync
      *
-     * @var string
+     * @var string|null
      */
     protected $httpStatus;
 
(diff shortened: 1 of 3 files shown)


skjnldsv (approved):

skjnldsv:
@DeepDiver1975 @evert who have review rights and/or who can we ping to try to get this forward? :)

susnux:
Fixed the unit tests (some existing ones were expecting incorrect data (status AND a propstat while only on of both is allowed).

susnux:
> who have review rights and/or who can we ping to try to get this forward? :)

Maybe @phil-davis ?

evert:
I don't see a problem with this change, and does seem like we got this wrong. I'm curious if you are running into any incompatibilities with specific clients.

skjnldsv:
> I don't see a problem with this change, and does seem like we got this wrong. I'm curious if you are running into any incompatibilities with specific clients.

We actually faced that issue here: https://github.com/nextcloud/server/pull/40474
Rewriting part of our frontend properly, we realized the backend was not responding accordingly :) 

We're using the [webdav-client](https://github.com/perry-mitchell/webdav-client) node library

PR #1495, opened 2023-10-25
fix: Handle `Depth` header for `COPY` as this is required by RFC

susnux:
Currently Sabre\DAV does not support the `Depth` header on `COPY` and `MOVE`.
But the RFC4918 requires servers to do so, the header is not required but:

* If the header is missing 'infinity' shall be assumed -> This is how we handle **every** request currently
* For `MOVE` the header, if available, **must** be 'infinity' -> We just have to check the header and throw bad request otherwise
* For `COPY` the header may 'infinity' or a positive integer including 0
  * The depths defines how many levels of children shall be copied
  * If the depth is zero only the collection and its properties are copied, but **not** any children

So especially the `COPY` method is currently broken, we noticed this when using the WebDAV litmus test bench (version 0.14) which checks for shallow copies (meaning copy a collection with depth 0).

This PR fixes all of this points, while trying to keep backwards compatible. Of cause any using application that implements `ICopyTarget::copyInto` is still broken and need to adapt to the depth change (it continues to work as before, but the bug fix does not propagate).

Reference from the RFC:
* The requirement on the server support of `COPY` `Depth` 'infinity' and '0': rfc4918#section-9.8.3
* For `MOVE` only `Depth: infinity` is allowed: rfc4918#section-9.9.2

### DIFF
--- a/lib/DAV/Server.php
+++ b/lib/DAV/Server.php
@@ -580,9 +580,10 @@ class Server implements LoggerAwareInterface, EmitterInterface
     /**
      * Returns the HTTP depth header.
      *
-     * This method returns the contents of the HTTP depth request header. If the depth header was 'infinity' it will return the Sabre\DAV\Server::DEPTH_INFINITY object
+     * This method returns the contents of the HTTP depth request header. If the depth header was 'infinity' it will return the Sabre\DAV\Server::DEPTH_INFINITY constant.
      * It is possible to supply a default depth value, which is used when the depth header has invalid content, or is completely non-existent
      *
+     * @param mixed $default default value to use if no header is set or has invalid value
      * @param mixed $default
      *
      * @return int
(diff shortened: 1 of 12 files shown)


lib/DAV/Server.php
```diff
@@ -725,10 +725,17 @@ public function getCopyAndMoveInfo(RequestInterface $request)
             throw new Exception\BadRequest('The destination header was not supplied');
         }
         $destination = $this->calculateUri($request->getHeader('Destination'));
-        $overwrite = $request->getHeader('Overwrite');
-        if (!$overwrite) {
-            $overwrite = 'T';
+
+        // Depth of inifinty is valid for MOVE and COPY. If it is not set the RFC requires to act like it was 'infinity'.
+        $depth = strtolower($request->getHeader('Depth') ?? 'infinity');
+        if ($depth !== 'infinity' && is_numeric($depth)) {
```
DeepDiver1975, line 731: In core plugin above it is stated that only `infinity` is allowed. Adding this logic here can reduce the necessary code changes a lot.

DeepDiver1975 (changes requested):

lib/DAV/Server.php
susnux: @DeepDiver1975 yes thats for `MOVE`, but for `COPY` we need to support also numeric values.

susnux:
ping again, also cc @phil-davis 

Would be great to get this fixed so sabre passes the litmus test v15 ( https://github.com/notroj/litmus previously http://www.webdav.org/neon/litmus/ )

susnux:
Anything I can help with to get this reviewed? (I do not want to bug you but this would be a great to get fixed :) ).

DeepDiver1975 (approved):

DeepDiver1975:
I miss the ci status - what's going on here?

DeepDiver1975:
> I miss the ci status - what's going on here?

@susnux can I ask you to rebase this pull request ... THX

susnux:
> can I ask you to rebase this pull request ... THX

@DeepDiver1975 sure, done!

susnux:
Noticed it is still open, is there anything I can help with?

ChristophWurst (approved):

kesselb:
cc @susnux can you rebase, to have a recent ci run?

susnux:
> can you rebase, to have a recent ci run?

Sure, done

susnux:
@kesselb I pushed a commit solving your comments

lib/DAV/Server.php
```diff
@@ -725,10 +726,18 @@ public function getCopyAndMoveInfo(RequestInterface $request)
...
-        $overwrite = $request->getHeader('Overwrite');
-        if (!$overwrite) {
-            $overwrite = 'T';
+
+        // Depth of infinity is valid for MOVE and COPY. If it is not set the RFC requires to act like it was 'infinity'.
+        $depth = $request->getHeader('Depth') ?? 'infinity';
+        if ('infinity' === strtolower($depth)) {
+            $depth = self::DEPTH_INFINITY;
+        } elseif (!ctype_digit($depth) || ((int) $depth) < 0) {
+            throw new Exception\BadRequest('The HTTP Depth header may only be "infinity", 0 or an positive integer');
```
alecpl: "a positive"

alecpl (approved):

provokateurin (approved):

Issue #1415, opened 2022-09-11
Release major version 5


phil-davis:
which will support PHP 7.4 8.0 8.1 8.2 and have type declarations on parameters and return values etc as far as possible with features available in PHP 7.4.

If possible, increase the phpstan level to 6. That requires adding a lot of parameters and return type detail in PHPdoc (specially of the shape of arrays. I will try adjusting a few things and then report how it goes, and how big a job it seems.

phil-davis:
Blocked by https://github.com/sabre-io/vobject/issues/588 - I need to work on that issue firts and get `sabre/vobject` 5.0.0 released. `sabre/dav` has a dependency on that.

phil-davis:
Status of `sabre` components as at 2022-09-11

Done:
cache - 2.0.0 (2022-08-23)
event - 6.0.0 (2022-08-29)
http - 6.0.0 (2022-08-31)
uri - 2.3.1 (2022-08-30) major version was not needed
xml - 3.0.0 (2022-08-17)

ToDo:
vobject - https://github.com/sabre-io/vobject/issues/588
dav - this issue

BinaryKitten:
hi - can i politely ask as to the status of the updates?

phil-davis:
> hi - can i politely ask as to the status of the updates?

I was making lots of progress, then got going on vobject https://github.com/sabre-io/vobject/pull/591 and it needs "a lot" (tm) of specifying types through the code. I have not got back to it for the last 3 months.

https://github.com/sabre-io/vobject/pull/591#issuecomment-1384864967

BinaryKitten:
> I was making lots of progress, then got going on vobject [sabre-io/vobject#591](https://github.com/sabre-io/vobject/pull/591) and it needs "a lot" (tm) of specifying types through the code. I have not got back to it for the last 3 months.
> 
> [sabre-io/vobject#591 (comment)](https://github.com/sabre-io/vobject/pull/591#issuecomment-1384864967)

Is there anything anyone can do to help?

DeepDiver1975:
Once #1497 is merged we can create a `4.x` branch and move towards 5.x on master .... objections @phil-davis @staabm ?

phil-davis:
yes, get as many little fixes and tweaks released in the 4.* series and then we can branch it, and do real work in master towards v5.

staabm:
sounds good to me

staabm:
[...]

PR #1502, opened 2023-11-09
feat: use php7.4 as minimum

DeepDiver1975:
refs #1415

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: [ '7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2' ]
+        php-versions: [ '7.4', '8.0', '8.1', '8.2' ]
         coverage: [ 'xdebug' ]
         streaming: [ false ]
         include:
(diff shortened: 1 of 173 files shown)


phil-davis:
@DeepDiver1975 there were 1807 unit tests running - for exampe https://github.com/sabre-io/dav/actions/runs/6862503131/job/18660341581 
But only 887 unit tests have run in gthis PR: https://github.com/sabre-io/dav/actions/runs/6856309480/job/18643234604?pr=1502

Maybe renaming various test file to `TestAbstract.php` has caused that?

DeepDiver1975:
I will look into this.

DeepDiver1975:
@phil-davis we have now 1640 tests. Previously test within sabre/http were executed as well - but they no longer exist in the composer package. This is fine form my pov as it is no the job of sabre/dav to run sabre/http tests

phil-davis:
> Previously test within sabre/http were executed as well
'
I had wondered why that was done, but without knowing the history I didn't like to stop running those sabre/http tests.

phil-davis (approved):
LGTM

phil-davis:
I will merge and we can progress from here towards a major version with type declarations etc.

PR #1504, opened 2023-11-09
fix: Add check for Lock-Token in request header.

DeepDiver1975:
rebased #981


DeepDiver1975:
word 2016 :zzz: -> close

Issue #1510, opened 2023-11-10
Bug, add principals programtically do not store display name and email.


ijortengab:
I want to auto create principals programatically. This the code:
```php
// @file index.php
use Sabre\DAV\MkCol;
$principalBackend = new Sabre\DAVACL\PrincipalBackend\PDO($pdo);
$path = 'principals/ijortengab';
$mkCol = new MkCol(["{DAV:}principal"], [
    '{DAV:}displayname' => 'IjorTengab',
    '{http://sabredav.org/ns}email-address' => 'ijortengab@example.org',
]);
$principalBackend->createPrincipal($path, $mkCol);
```

Expected result: Add new record inside `principals` table, with column `uri` fill value `principals/ijortengab`, column `email` fill value `ijortengab@example.org`, and column `displayname` fill value `IjorTengab`.

Actual result: Record added, but column `email`  is NULL, and column `displayname` is NULL.

I've found the solution: inside lib\DAVACL\PrincipalBackend\PDO.php inside `updatePrincipal` method, append this line 
```
$propPatch->commit();
```

DeepDiver1975:
`commit()` is expected to be called on top level to perform all operations in one transaction.

All you are missing in your script is:
```php
$mkCol->commit();

PR #1511, opened 2023-11-10
Bugfix #1510, fill correctly column email and displayname inside prin…

ijortengab:
…cipals table.


DeepDiver1975:
see #1510 for explanation

PR #1514, opened 2023-11-22
fix: restore autoPrefix property of Href

phil-davis:
use of this property in code was removed in PR
https://github.com/sabre-io/dav/pull/801
But the property is left here because old data may still exist that has this property saved.

See discussion in https://github.com/sabre-io/Baikal/issues/1154

Fixes https://github.com/sabre-io/Baikal/issues/1154 and related issue #1445

### DIFF
--- a/lib/DAV/Xml/Property/Href.php
+++ b/lib/DAV/Xml/Property/Href.php
@@ -35,6 +35,19 @@ class Href implements Element, HtmlOutput
     protected $hrefs;
 
     /**
+     * Automatically prefix the url with the server base directory.
+     * Note: use of this property in code was removed in PR:
+     * https://github.com/sabre-io/dav/pull/801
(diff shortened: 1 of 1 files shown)


phil-davis:
If this is the accepted fix, then I will backport it to 4.5 and make a 4.5.1 release.

phil-davis:
> Was is always protected? I don't know right now if unserialize has issues with property access.....

I copied from what was deleted in:
https://github.com/sabre-io/dav/pull/801/files#diff-abad1406d5405d56e66aa47e67d2277a2707befef8d570f871a2c243676729b0

DeepDiver1975:
just looked it up - serialize/unserialize require the access level to be unchanged.
Since this is the case I think we can safely merge this.

phil-davis:
Backported to 4.5 branch in PR #1515

PR #1518, opened 2023-11-24
feat: add propFindUnfiltered public method to Client

phil-davis:
1) add an extra unit test to cover the existing behaviour of `propFind` when the response has multiple resources listed, and the resources have different status.

2) refactor out the common code that we want to keep using into `doPropFind`

3) add `propFindUnfiltered` method that returns all the propfind data, including for resources that had a non-200 status, and a unit test to cover it.

4) return both properties and status in propFindUnfiltered
The propfind data was returned (key-value array of properties and their values). But the individual statuses were not anywhere in the returned data. The caller is going to want to be able to find out the status of the propfind of each resource. Where to put that in the returned data, without polluting the key-values of the properties themselves? I have made the returned array have:
- array of resources keyed by the resource name
- inside that, and array with keys 'properties' and 'status'
- 'properties' has the key-value array of properties and their values
- 'status' has the status returned for that resource

Question: I have currently assumed that there is just one status value for each resource - for example "file1.txt" might have got a 418 "I'm a teapot" status, and there are a list of properties that were returned. But is it possible that each property might have a different status? 
(some properties were "easy" for the server to get, and have 200 status, other properties were "secure" and had a 403 "forbidden" status for the requesting user, other properties got 404 "not found" because that file in the folder did not have that meta-data...)

PR #1519, opened 2023-11-27
feat: return PROPFIND properties indexed by status

phil-davis:
This is on top of PR #1518 

See the latest commit - it returns the PROPFIND data keyed by status. All the items with status `200` are in an array with key `200`, all the items with status `400` are in an array with key `400`.

That simplifies the code a bit, and avoids having the keywords "properties" and "status'.

I'm not sure if this is a better way or not.

### DIFF
--- a/lib/DAV/Client.php
+++ b/lib/DAV/Client.php
@@ -174,7 +174,7 @@ class Client extends HTTP\Client
     }
 
     /**
-     * Does a PROPFIND request.
+     * Does a PROPFIND request with filtered response.
      *
      * The list of requested properties must be specified as an array, in clark
      * notation.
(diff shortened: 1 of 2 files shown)


individual-it:
This works fine for the ocis-php-sdk, see https://github.com/owncloud/ocis-php-sdk/pull/117

phil-davis:
> This works fine for the ocis-php-sdk, see [owncloud/ocis-php-sdk#117](https://github.com/owncloud/ocis-php-sdk/pull/117)

@individual-it then are you happy that we can go with this data structure being returned?

individual-it:
@phil-davis yes, if this can be merged and release it will help me a lot

lib/DAV/Client.php
```diff
@@ -174,27 +174,107 @@ public function __construct(array $settings)
...
+     * made to the server to also return all child resources.
+     *
+     * The returned array will contain a list of resources as keys and
+     * a multi-level array containing status and properties as value.
+     *
+     * The multi-level array of status and properties is formatted the same as what is
+     * documented for parseMultiStatus.
+     *
+     * @param string $url
+     * @param int    $depth
```
staabm, line 273:
```suggestion
     * @param 1|0 $depth
```

lib/DAV/Client.php
```diff
@@ -174,27 +174,107 @@ public function __construct(array $settings)
...
+     * and an array of properties as values.
      *
-     * The properties array will contain the list of properties. Only properties
-     * that are actually returned from the server (without error) will be
+     * The array of properties will contain the properties as keys with their values as the value.
+     * Only properties that are actually returned from the server without error will be
      * returned, anything else is discarded.
      *
+     * @param string $url
+     * @param int    $depth
```
staabm, line 195:
```suggestion
     * @param 0|1 $depth
```

lib/DAV/Client.php
```diff
@@ -174,27 +174,107 @@ public function __construct(array $settings)
...
+     * For depth 1, the returned array will contain a list of resources as keys and
+     * a multi-level array containing status and properties as value.
+     *
+     * The multi-level array of status and properties is formatted the same as what is
+     * documented for parseMultiStatus.
+     *
+     * All properties that are actually returned from the server are returned by this method.
+     *
      * @param string $url
      * @param int    $depth
```
staabm, line 239:
```suggestion
     * @param 1|0 $depth
```

lib/DAV/Client.php
```diff
@@ -174,27 +174,107 @@ public function __construct(array $settings)
...
+     *
+     * The multi-level array of status and properties is formatted the same as what is
+     * documented for parseMultiStatus.
+     *
+     * @param string $url
+     * @param int    $depth
+     *
+     * @return array
+     */
+    private function doPropFind($url, array $properties, $depth = 0)
```
staabm, line 277: can/should we have native return types?

lib/DAV/Client.php
```diff
@@ -174,27 +174,107 @@ public function __construct(array $settings)
     }
 
     /**
-     * Does a PROPFIND request.
+     * Does a PROPFIND request with filtered response.
```
staabm, line 177: maybe we should leave a word what/why is it filtered

lib/DAV/Client.php
phil-davis: I added parameter and return types.
We can do both here, because `master`has code towards doing this across the codebase (for a future major release).

I will have to be more careful when backporting, to make sure that existing public methods are correctly backward-compatible.

lib/DAV/Client.php
phil-davis: added some words

lib/DAV/Client.php
phil-davis: done - when backporting I will need to see if I can also do this in the "old" releases.

lib/DAV/Client.php
phil-davis: done

phil-davis:
@staabm please review again

PR #1535, opened 2024-01-23
chore(docs): change links to https if possible

Rotzbua:


phil-davis (approved):
Those all work with https, so might as well use it.

PR #1536, opened 2024-01-23
chore(docs): fix typos

Rotzbua:
Just in documentation.

### DIFF
--- a/lib/CalDAV/Notifications/Node.php
+++ b/lib/CalDAV/Notifications/Node.php
@@ -70,7 +70,7 @@ class Node extends DAV\File implements INode, DAVACL\IACL
     /**
      * Returns the etag for the notification.
      *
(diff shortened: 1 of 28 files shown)


lib/CalDAV/Notifications/Node.php
```diff
@@ -101,7 +101,7 @@ public function delete()
     /**
      * Returns the owner principal.
      *
-     * This must be a url to a principal, or null if there's no owner
+     * This must be an url to a principal, or null if there's no owner
```
phil-davis, line 104: I guess there is "discussion" about "an url" or "a URL", when pronouncing each letter then it is "a"
"a U, an R and an L". But I will leave this change here - it does not matter either way.

phil-davis (approved):
Thanks, I obviously haven't run the spell-checker over this repo!

PR #1537, opened 2024-01-26
fix(docs): add new php requirements

Rotzbua:
Main branch has newer php requirement than release.


phil-davis (approved):
Thanks for noticing

PR #1542, opened 2024-02-07
Fixed issue with multiple invites being send out with a comma in the recipient name/address

wlamers:
When using for example Thunderbird to create an invite an adding Attendees that have a comma in their name in the addressbook (of from collected emailaddresses in received emails) like e.g.:

"test, test <test@test.com>"

multiple RCPT TO: commands are send over SMTP, potentially causing issues with the mail server (postfix/dovecot). In our case the mail server want to send the invite to both the user "test" as the email "test@test.com". User "test" does not exist, causing the invite to end up in the 'catch-all' email account.

By extracting the emailadress using a regular expression the issue is resolved.


phil-davis:
@wlamers are you able to add unit tests?

PR #1545, opened 2024-02-20
fix getNodeForPath caching

icewind1991:
The change https://github.com/sabre-io/dav/pull/1060 from the recursive to iterative `getNodeForPath` means that any cached parent node isn't used anymore.

This adds some logic to find the nearest parent that is cached and start the iterative logic from there.

### DIFF
--- a/lib/DAV/Tree.php
+++ b/lib/DAV/Tree.php
@@ -62,8 +62,20 @@ class Tree implements INodeByPath
             return $this->rootNode;
         }
 
-        $parts = explode('/', $path);
         $node = $this->rootNode;
+
+        // look for any cached parent and collect the parts below the parent
+        $parts = [];
+        $remainingPath = $path;
+        do {
+            list($remainingPath, $baseName) = Uri\split($remainingPath);
+            array_unshift($parts, $baseName);
+
+            if (isset($this->cache[$remainingPath])) {
+                $node = $this->cache[$remainingPath];
+                break;
+            }
+        } while ('' !== $remainingPath);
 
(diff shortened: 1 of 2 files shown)


patrickTUD:
@icewind1991 - you rock! See https://github.com/nextcloud/server/issues/42143

ingeniumnrsp:
For those that don't have root access to their NC installs, which version can we expect the fix to arrive in?

Thanks!

lib/DAV/Tree.php
```diff
@@ -62,9 +62,21 @@ public function getNodeForPath($path)
             return $this->rootNode;
         }
 
-        $parts = explode('/', $path);
         $node = $this->rootNode;
 
+        // look for any cached parent and collect the parts below the parent
+        $parts = [];
+        $remainingPath = $path;
```
phil-davis, line 69: Note: we know that `$path` is not the empty string, because of the check a few lines above.
So the call to `Uri\split()` below will always pass a non-empty string.
So we don't have to worry about the fact that `Uri\split` returns `[null, null]` when passed and empty string.

Just noting this, because I saw that `Uri\split` can return `[null, null]` and I was concerned about that - it would cause it to be called again with `null` and explode.

phil-davis (approved):
LGTM

phil-davis:
@staabm @DeepDiver1975 any comment? or merge and release?

DeepDiver1975 (approved):

phil-davis:
Released in https://github.com/sabre-io/dav/releases/tag/4.7.0

PR #1553, opened 2024-07-05
DAV\PropPatch::handleRemaining should only register one callback

paulmhh:
currently handleRemaining() will create one callback for each remaining property update but with the whole array of remaining properties as parameter. Thus doCallBackMultiProp() is being used (in each callback! It should be either one callback with an array as parameter, or it should multiple callbacks with a string as parameter)

### DIFF
--- a/lib/DAV/PropPatch.php
+++ b/lib/DAV/PropPatch.php
@@ -123,12 +123,12 @@ class PropPatch
         foreach ($properties as $propertyName) {
             // HTTP Accepted
             $this->result[$propertyName] = 202;
-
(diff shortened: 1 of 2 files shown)


DeepDiver1975 (changes requested):
would be great to have a unit test for this - thx

paulmhh:
sorry for the force pushes, had some trouble to run the github workflow and then the cs-fixer wanted some minor change and I had to iterate.

I did test the test without my fix. It fails. With the fix applied it is green.

alecpl (approved):

alecpl:
@phil-davis ping, this looks ready.

PR #1559, opened 2024-08-21
Allow for custom iTipBroker

SebastianKrupinski:
Uses a protected function to generate the iTipBroker so that is can be overriden and used to generate a custom broker class.

### DIFF
--- a/lib/CalDAV/Schedule/Plugin.php
+++ b/lib/CalDAV/Schedule/Plugin.php
@@ -25,6 +25,7 @@ use Sabre\HTTP\ResponseInterface;
 use Sabre\VObject;
 use Sabre\VObject\Component\VCalendar;
 use Sabre\VObject\ITip;
+use Sabre\VObject\ITip\Broker;
 use Sabre\VObject\ITip\Message;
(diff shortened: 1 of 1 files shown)


lib/CalDAV/Schedule/Plugin.php
```diff
@@ -994,4 +994,14 @@ public function getPluginInfo()
             'link' => 'http://sabre.io/dav/scheduling/',
         ];
     }
+
+    /**
+     * Returns an instance of the iTip\Broker
+     * 
+     * @return Broker
```
ChristophWurst, line 1001: Needs an important and please use a return type hint

ChristophWurst: Looks good now!

ChristophWurst:
@phil-davis I hope this change makes sense. We add the protected factory method so that sub classes of the plugin can work with a sub-classed broker as well.

SebastianKrupinski:
@DeepDiver1975 Done.

DeepDiver1975 (approved):

NameNoQuality (approved):

PR #1561, opened 2024-09-15
refactor: Refactor ternary to elvis operator where possible

ChristophWurst:
This should make the code a bit easier to read.

The Elvis operator is available since PHP5.3.


phil-davis (approved):
LGTM

PR #1562, opened 2024-09-15
refactor: Change class strings to ::class constants

ChristophWurst:
This is a scripted refactoring to replace class strings with ::class constant references.

I chose not to import the classes to keep the diff clean.

### DIFF
--- a/tests/Sabre/DAVACL/Xml/Property/ACLTest.php
+++ b/tests/Sabre/DAVACL/Xml/Property/ACLTest.php
@@ -12,7 +12,7 @@ class ACLTest extends \PHPUnit\Framework\TestCase
     public function testConstruct()
     {
         $acl = new Acl([]);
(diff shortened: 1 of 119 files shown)


tests/Sabre/DAVACL/Xml/Property/ACLTest.php
```diff
@@ -12,7 +12,7 @@ class ACLTest extends \PHPUnit\Framework\TestCase
     public function testConstruct()
     {
         $acl = new Acl([]);
-        self::assertInstanceOf('Sabre\DAVACL\Xml\Property\ACL', $acl);
+        self::assertInstanceOf(\Sabre\DAVACL\Xml\Property\Acl::class, $acl);
```
ChristophWurst, line 15: ^ small casing issue revealed by the refactoring :)

phil-davis (approved):

PR #1568, opened 2024-10-29
Allow for custom iTipBroker (#1559) [4.6]

phil-davis:
Backport #1559 to 4.6

### DIFF
--- a/lib/CalDAV/Schedule/Plugin.php
+++ b/lib/CalDAV/Schedule/Plugin.php
@@ -25,6 +25,7 @@ use Sabre\HTTP\ResponseInterface;
 use Sabre\VObject;
 use Sabre\VObject\Component\VCalendar;
 use Sabre\VObject\ITip;
+use Sabre\VObject\ITip\Broker;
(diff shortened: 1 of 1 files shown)


kesselb:
Hi @phil-davis,

I was wondering with @SebastianKrupinski if you want to do a release for 4.6 with the patch? ;)

phil-davis:
> I was wondering with @SebastianKrupinski if you want to do a release for 4.6 with the patch?

This code ended up being released in https://github.com/sabre-io/dav/releases/tag/4.7.0 because it is an extra "feature" (along with other things in 4.7.0 that could be considered "features" rather than bug fixes.)

@kesselb 4.7.0 still supports all PHP 7.1 through 8.4. Is there any reason that you can't use 4.7.0 ?

SebastianKrupinski:
Hi @phil-davis,

The patch allows us to fix some bugs in NC, and our supported stable releases use 4.6, so a updated 4.6 release would allow us to fix the issues, without having to custom patch those stable releases.

Thank you

PR #1591, opened 2025-07-25
Uri in xml response must be url encoded.

flymarq:
The Uri is derived from the request with the method getPath, which generates a valid path fragment suitable for file and directory names.

In the XML-response to a lock-request the uri is used for lockroot element. If the uri contains a space or german umlauts, a strict client like libneon, which is used in davfs2, refuses the lock reponse and files cannot be copied.

### DIFF
--- a/lib/DAV/Locks/Plugin.php
+++ b/lib/DAV/Locks/Plugin.php
@@ -234,6 +234,7 @@ class Plugin extends DAV\ServerPlugin
         }
 
         $this->lockNode($uri, $lockInfo);
+        $lockInfo->uri = urlencode($lockInfo->uri);
 
         $response->setHeader('Content-Type', 'application/xml; charset=utf-8');
         $response->setHeader('Lock-Token', '<opaquelocktoken:'.$lockInfo->token.'>');
--- a/tests/Sabre/DAV/Locks/PluginTest.php
+++ b/tests/Sabre/DAV/Locks/PluginTest.php
@@ -920,4 +920,14 @@ class PluginTest extends DAV\AbstractServerTestCase
         $this->server->httpRequest = $request;
         $this->locksPlugin->getTimeoutHeader();
     }
+
+    public function testLockWithSpaces()
+    {
+        $request = new HTTP\Request('LOCK', '/test .txt', [
+            'Timeout' => 'second-5, infinite',
+        ]);
+        $this->server->httpRequest = $request;
+        $this->server->exec();
+        self::assertEquals(201, $this->response->status);
+    }
 }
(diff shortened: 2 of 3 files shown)


alecpl:
I see the second commit here is unrelated and should be removed. Also, it would be great to have a test added in tests/Sabre/DAV/Locks/PluginTest.php.

flymarq:
My fault, the second commit does not belong to this issue.

I can try to add a test, but I'm afraid it won't work unless a strict client like libneon performs the test.

tests/Sabre/DAV/Locks/PluginTest.php
```diff
@@ -920,4 +920,14 @@ public function testGetTimeoutHeaderInvalid()
...
     }
+
+    public function testLockWithSpaces()
+    {
+        $request = new HTTP\Request('LOCK', '/test .txt', [
+            'Timeout' => 'second-5, infinite',
+        ]);
+        $this->server->httpRequest = $request;
+        $this->server->exec();
+        self::assertEquals(201, $this->response->status);
```
alecpl: This test should assert the activelock/lockroot/href property in the XML response. That's where the encoding problem would be visible.

alecpl (changes requested):

lib/DAV/Locks/Plugin.php
```diff
@@ -234,6 +234,7 @@ public function httpLock(RequestInterface $request, ResponseInterface $response)
         }
 
         $this->lockNode($uri, $lockInfo);
+        $lockInfo->uri = urlencode($lockInfo->uri);
```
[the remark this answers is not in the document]
flymarq: I do not have any problmes or errors in PROPFIND calls, why should I change anything?

I can create directories with spaces in the name and inside this directory I can create a file with spaces in the name.

From my experience only LOCK is broken/affected.

tests/Sabre/DAV/Locks/PluginTest.php
flymarq: My starting point is this line in the logs of the webserver:
`"LOCK /F%20G/F%20G%20H%20J.txt HTTP/1.1" 423 327 "-" "davfs2/1.7.2 neon/0.35.0"`

and on the terminal I get 
`touch: '/home/flynn/media/flynnux/F G/F G H J.txt' kann nicht berührt werden: Eingabe-/Ausgabefehler`

In the error case the XML-response contains an exception: 
`<s:exception>Sabre\DAV\Exception\Conflict no-conflicting-lock>`

On success the the token:
 `<d:locktoken><d:href>opaquelocktoken:c546f02c-5fe7-4a9f-bf78-3bd7c57ef72a</d:href></d:locktoken>`
 
 This is, what I try test, tell me if I'm wrong ...

alecpl: ConflictingLock exceptions are thrown in line 170 and 178. Your change is after that, so I don't see how that fixes anything, especially after `lockNode()`.

PR #1599, opened 2025-10-10
feat(carddav): allow pushing down filtering of reports

ChristophWurst:
A `addressbook-query` currently loads the full address book to filter in memory. Depending on the implementation it should be possible to push some of the filtering down to the implementation, e.g. to filter on a DB level first. This improves memory footprint and performance drastically.

I'm opening this as proof of concept and hope to get some input on the approach and whether this would be accepted.

The same mechanism exists for caldav, where I took inspiration from. For BC I chose to not add a new method to the carddav backend interface.

https://github.com/nextcloud/server/pull/55669 shows what it will look like in an implementation.

### DIFF
--- a/lib/CardDAV/IAddressBookObjectContainer.php
+++ b/lib/CardDAV/IAddressBookObjectContainer.php
@@ -1,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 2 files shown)


ChristophWurst:
@phil-davis friendly ping

PR #1612, opened 2025-12-09
Use Uri\Split() instead of basename()

### LINKED ISSUE #1348: Purge "basename" from everything
basename() is not multibyte-safe and seems to have been removed from almost everywhere, but it still lingers in
\sabre\dav\bin\migrateto30.php
and is lurking in a comment in
\sabre\dav\lib\DAV\Browser\Plugin.php

Also, it is used in the docs, for example in
https://sabre.io/dav/virtual-filesystems/
which in this case bit me -- I had no idea basename was even a native PHP function, and its undefined utf8 behavior if not prepended with set_locale, and so it dropped the first mb char of every file name if present, which was unexpected. Easily fixed with Uri\split but that should be used everywhere for clarity I think.

alecpl:
Fixes #1348.

### DIFF
--- a/bin/migrateto30.php
+++ b/bin/migrateto30.php
@@ -149,7 +149,7 @@ $stmt1 = $pdo->prepare('INSERT INTO propertystorage (path, name, valuetype, valu
 while ($row = $result->fetch(\PDO::FETCH_ASSOC)) {
     // Inserting the new record
     $stmt1->execute([
-        'addressbooks/' .  \Sabre\Uri\split($row['uri'])[1],
+        'addressbooks/'.\Sabre\Uri\split($row['uri'])[1],
         '{http://calendarserver.org/ns/}me-card',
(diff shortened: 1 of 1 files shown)


bin/migrateto30.php
```diff
@@ -149,7 +149,7 @@
 while ($row = $result->fetch(\PDO::FETCH_ASSOC)) {
     // Inserting the new record
     $stmt1->execute([
-        'addressbooks/'.basename($row['uri']),
+        'addressbooks/'.\Sabre\Uri\split($row['uri'])[1],
```
ChristophWurst: Are these URIs always up to two parts? It could make sense to use something like `list(, $basename) = \Sabre\Uri\split($row['uri'])`. It's used like that in other section of the codebase.

alecpl: I mean, it does not really make sens (nor change anything) to use list here.

ChristophWurst (approved):

PR #1614, opened 2025-12-11
Check node type before getOwner() call

### LINKED ISSUE #1345: \Sabre\CalDAV\Schedule\Plugin::beforeUnbind calls \Sabre\DAVACL\IACL::getOwner without a type check
https://github.com/sabre-io/dav/blob/c1afdc77a95efea6ee40c03c45f57c3c0c80ec22/lib/CalDAV/Schedule/Plugin.php#L389

The node might not have ACLs implemented. In that case this calls a method that doesn't exist.

In my case it's an `\Sabre\CalDAV\ICalendarObject` that gets deleted. Those often have ACLs, but not always. It might make sense to exit early when IACL is not implemented by the node.

alecpl:
Fixes #1345

### DIFF
--- a/lib/CalDAV/Schedule/Plugin.php
+++ b/lib/CalDAV/Schedule/Plugin.php
@@ -378,7 +378,7 @@ class Plugin extends ServerPlugin
 
         $node = $this->server->tree->getNodeForPath($path);
 
-        if (!$node instanceof ICalendarObject || $node instanceof ISchedulingObject) {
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Looks good! Thanks

alecpl:
I'm not sure how to reproduce or add a test for this case. So, someone else would have to do this.

alecpl:
@phil-davis ping, this is an easy fix, but I'm unable to provide a test.

PR #1617, opened 2025-12-23
Don't include quota properties in allprop request

### LINKED ISSUE #1616: A <DAV:allprop> PROPFIND request SHOULD NOT return quota properties
It looks to me that when I implement `IQuota` interface on a collection `quota-used-bytes` and `quota-available-bytes` get included in a PROPFIND request with just the `<DAV:allprop/>` element.

This is in violation of https://datatracker.ietf.org/doc/html/rfc4331#section-2

alecpl:
Fixes #1616


miaulalala (approved):

alecpl:
@phil-davis ping, this one is pretty obvious and documented, but will require a decision.

Issue #917, opened 2016-11-29
Postgresql Backend: invalid input syntax for type bytea


tobiastom:
I'm trying to store a calendar event. The macOS client Calendar receives a 500 server error shows that it had a problem in process `CalDAVWriteEntityQueueableOperation`.

I did install a proxy to investigate the data the is transmitted, and I can see that the server responds with this XML data:

    <?xml version="1.0" encoding="utf-8"?>
    <d:error xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
      <s:sabredav-version>3.2.0</s:sabredav-version>
      <s:exception>PDOException</s:exception>
      <s:message>SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type bytea</s:message>
    </d:error>

There is not error inside the PHP error log. Unfortunately I cannot provide the full event data and for now I'm not able to reproduce it with a minified example. This is somehow a censored request:

	BEGIN:VCALENDAR
	VERSION:2.0
	PRODID:-//Apple Inc.//Mac OS X 10.12.1//EN
	CALSCALE:GREGORIAN
	BEGIN:VEVENT
	TRANSP:OPAQUE
	DTEND:20161205T151500Z
	LAST-MODIFIED:20161123T115407Z
	ORGANIZER;CN="XXXXXX";SCHEDULE-AGENT=CLIENT:mailto:XXXXXXXXX@XXXXXXXXXXXX
	 XXXXXXXXX
	UID:XXXXXXXXXXXXXXXXXXXXXXXXXXX@XXXXXXXXX
	DTSTAMP:20161129T161928Z
	LOCATION:XXXXXXXXX
	DESCRIPTION:XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
	 XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
	STATUS:CONFIRMED
	SEQUENCE:0
	SUMMARY:XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
	DTSTART:20161205T143000Z
	X-APPLE-TRAVEL-ADVISORY-BEHAVIOR:AUTOMATIC
	CREATED:20161123T114949Z
[...]

PR #1623, opened 2026-02-09
Fix PDO::PARAM_LOB binding for BYTEA columns in CalDAV and CardDAV

### LINKED ISSUE #1587: PostgreSQL error caused by comma
Sabre/DAV 4.7.0
PostgreSQL 15

Whenever a calendar or contact field contains a comma `,` it will fail to save.

FYI, a comma is escaped as `\,`. The backslash escaped as `\\` does not cause any problem.

```
  <s:sabredav-version>4.7.0</s:sabredav-version>
  <s:exception>PDOException</s:exception>
  <s:message>SQLSTATE[22P02]: Invalid text representation: 7 ERROR:  invalid input syntax for type bytea
CONTEXT:  unnamed portal parameter $3 = '...'</s:message>
</d:error>
```

sylvinus:
Fixes #1587 and #917

## Problem

On PostgreSQL, inserting iCalendar or vCard data containing backslash escape sequences (e.g. `\,` or `\n`, which are standard RFC 5545 TEXT escapes) into the `calendardata` or `carddata` BYTEA columns fails with:

      SQLSTATE[22P02]: Invalid text representation: 7 ERROR:
      invalid input syntax for type bytea

This happens because PDO binds parameters as `PARAM_STR` by default. For BYTEA columns, PostgreSQL's text input parser then interprets backslash sequences as invalid octal escapes.

## Fix

Bind `calendardata`/`carddata` parameters with `PDO::PARAM_LOB` in all write methods, consistent with the existing fix in `Sabre\DAV\PropertyStorage\Backend\PDO::propPatch()`. Also switch to named parameters with `bindParam()` and explicit types for all parameters, matching the PropertyStorage style.

Read methods now handle the `is_resource()` case since PDO may return BYTEA data as a PHP stream.

## Test

Added `testCreateCalendarObjectWithIcsEscapes` in `AbstractPDOTestCase` — inserts ICS data with `\,` and `\n` escapes, verifies round-trip integrity. Passes on PostgreSQL with the fix, fails without it.

I can also confirm this fixes our own codebase where we had large ICS import break 10% of the time because of such characters in events.

### DIFF
--- a/lib/CalDAV/Backend/PDO.php
+++ b/lib/CalDAV/Backend/PDO.php
@@ -466,7 +466,7 @@ SQL
             'lastmodified' => (int) $row['lastmodified'],
             'etag' => '"'.$row['etag'].'"',
             'size' => (int) $row['size'],
-            'calendardata' => $row['calendardata'],
+            'calendardata' => is_resource($row['calendardata']) ? stream_get_contents($row['calendardata']) : $row['calendardata'],
             'component' => strtolower($row['componenttype']),
          ];
     }
@@ -507,7 +507,7 @@ SQL
                     'lastmodified' => (int) $row['lastmodified'],
                     'etag' => '"'.$row['etag'].'"',
                     'size' => (int) $row['size'],
-                    'calendardata' => $row['calendardata'],
+                    'calendardata' => is_resource($row['calendardata']) ? stream_get_contents($row['calendardata']) : $row['calendardata'],
                     'component' => strtolower($row['componenttype']),
                 ];
             }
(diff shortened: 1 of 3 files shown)


lib/CalDAV/Backend/PDO.php
```diff
@@ -466,7 +466,7 @@ public function getCalendarObject($calendarId, $objectUri)
             'lastmodified' => (int) $row['lastmodified'],
             'etag' => '"'.$row['etag'].'"',
             'size' => (int) $row['size'],
-            'calendardata' => $row['calendardata'],
+            'calendardata' => is_resource($row['calendardata']) ? stream_get_contents($row['calendardata']) : $row['calendardata'],
```
alecpl: The fact that you use LOB on INSERT does not mean that it will become a resource in SELECT. Imo, this part of the PR does not make sense. Or do I miss something?

provokateurin (approved):

PR #1631, opened 2026-03-22
Examples correction

buckaroo-labs:
change <code>$server->start();</code> to <code>$server->exec();</code>; provide MySQL example code


buckaroo-labs:
I would rather have split this into 2 requests, one for the examples correction and one for the year-2038 fixes. Still learning about how pull requests work.

PR #1633, opened 2026-03-26
chore(deps): Update phpstan

provokateurin:


ChristophWurst (approved):
protected -> private seems safe because it's just the test files 👍

PR #1635, opened 2026-03-29
#1588: Allow negative linux epoch

### LINKED ISSUE #1588: Linux Epoch prevents events (birthdays) being created older than 1st January 1970
# Preamble

This bug was issue (sabre-io/Baikal#1343) was initially discovered in Baikal. Attempting to import ICS file containing birthday's which were pre-1970 (Linux Epoch) would throw an error with the Postgres database regarding constraints violation. The `firstOccurrence` value would be set to `0` when the `getTimestamp()` was negative. In the database this value needed to be greater than 0 resulting in a constraint violation. The issue is minor and can be remedied easily with a modification of the constraint to accept values equal to and greater than 0. Please read the issue referenced above for more details.

# Root cause

The issue was traced to this library in the following lines:
https://github.com/sabre-io/dav/blob/7183a67c4c0f890be7d5ceec5331435db1987186/lib/CalDAV/Backend/PDO.php#L636-L668

Although the issue is niche calendars *should* be allowed to store events that have their first occurrence before the Linux Epoch.

abpai94:
Closes #1588

### DIFF
--- a/lib/CalDAV/Backend/PDO.php
+++ b/lib/CalDAV/Backend/PDO.php
@@ -674,14 +674,6 @@ SQL
                     $lastOccurence = $end->getTimeStamp();
                 }
             }
-
-            // Ensure Occurence values are positive
-            if ($firstOccurence < 0) {
-                $firstOccurence = 0;
-            }
-            if ($lastOccurence < 0) {
-                $lastOccurence = 0;
-            }
         }
 
         // Destroy circular references to PHP will GC the object.
--- a/lib/CalDAV/Backend/PDO.php
+++ b/lib/CalDAV/Backend/PDO.php
@@ -674,6 +674,12 @@ SQL
                     $lastOccurence = $end->getTimeStamp();
                 }
             }
+
+            // Ensure Occurence values are positive
+             if (date_diff($firstOccurence, $lastOccurence)->invert) {
+                 throw new \Sabre\DAV\Exception('Calendar firstOccurence and lastOccurence must be consecutive');
+             }
+
         }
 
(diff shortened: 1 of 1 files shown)


abpai94:
Possibly a bit overkill to verify that `firstOccurence` and `lastOccurence` are consecutive considering as 64 bit INT timestamps will roll-over in the year 2262. But if there are still people running 32 bit systems in 2038 maybe it will come in handy.

ralflang:
> Possibly a bit overkill to verify that `firstOccurence` and `lastOccurence` are consecutive considering as 64 bit INT timestamps will roll-over in the year 2262. But if there are still people running 32 bit systems in 2038 maybe it will come in handy.

I feel different about this. People may very well use their calendaring/event system to model zany things such as the repetition of easter or moon phases or anniversaries of battles because it contains all the complicated date math already - and exposing them to caldav/carddav may be an unintended but welcome side effect. Being defensive about dates outside our narrow PIM horizon isn't overkill, it's prudent.

abpai94:
@ralflang yes you make a good point, we should cover the edge cases and throw an error if we ever end up rolling over to prevent dates getting wrongly added to the calendar.

However I used https://www.epochconverter.com/ to calculate the 64 bit date (year 2262) and I **think** the website didn't provide the right answer if I understood correctly. But I'm still working on verifying this before pushing for a fix.

abpai94:
@ralflang as a heads up you may want to checkout the following issues which is also related sabre-io/Baikal#1344 where I have proposed some changes to the MySQL and Postgres database schemas. If you are using either of them you should apply the patch I have proposed.

PR #1637, opened 2026-04-06
fix(CalDAV/Backend/PDO): bind transparent as int in updateCalendar

sylvinus:
`updateCalendar` stored the `transparent` field as a PHP boolean, which the pgsql PDO driver serializes as the empty string `''` when the value is `false`. The `transparent SMALLINT NOT NULL` column then rejects the parameter with SQLSTATE 22P02.

This looks like a follow-up, 10 years later (!), of #816 :-)


sylvinus:
@provokateurin could you have a look? We'd love to switch back to mainline in our project. Thanks a lot!!

provokateurin (approved):

PR #1641, opened 2026-06-09
Allow passing calendar-user-type value from PrincipalBackend.

dehart:
Enhanced principal calendar-user-type property handle to allow passing a value from the PrincipalBackend instead of always default as INDIVIDUAL allowed values are RESOURCE, ROOM, GROUP and INDIVIDUAL


phil-davis:
@dehart please provide some additional unit tests do demonstrate the new behavior.

PR #1661, opened 2026-07-09
Add some type annotations

CarlSchwan:
This has been sitting in a local stash for a while

### DIFF
--- a/lib/CalDAV/Schedule/Plugin.php
+++ b/lib/CalDAV/Schedule/Plugin.php
@@ -314,7 +314,7 @@ class Plugin extends ServerPlugin
      * @param mixed             $modified     the iCalendar object has been touched
      * @param mixed             $isNew        Whether this was a new item or we're updating one
      */
-    public function calendarObjectChange(RequestInterface $request, ResponseInterface $response, VCalendar $vCal, $calendarPath, &$modified, $isNew)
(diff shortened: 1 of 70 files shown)


lib/CalDAV/Schedule/Plugin.php
```diff
@@ -316,7 +316,7 @@ public function propPatch($path, PropPatch $propPatch)
      * @param mixed             $modified     the iCalendar object has been touched
      * @param mixed             $isNew        Whether this was a new item or we're updating one
      */
-    public function calendarObjectChange(RequestInterface $request, ResponseInterface $response, VCalendar $vCal, $calendarPath, &$modified, $isNew)
+    public function calendarObjectChange(RequestInterface $request, ResponseInterface $response, VCalendar $vCal, $calendarPath, & $modified, $isNew)
```
ChristophWurst: unrelated?

ChristophWurst (commented):
Looks good

Reference args formatting could be split into a separate CS clean-up IMO

lib/CalDAV/Schedule/Plugin.php
CarlSchwan: it's my local cs-fixer doing that and on the CI it wants the original form back :( This is why I create the PR to update cs-fixer

phil-davis: php-cs-fixer latest version has reverted those back (space removed)

phil-davis (approved):

PR #1662, opened 2026-07-09
Update php-cs-fixer

CarlSchwan:
From the v2 which is EOL to v3

### DIFF
--- a/examples/groupwareserver.php
+++ b/examples/groupwareserver.php
@@ -32,7 +32,7 @@ date_default_timezone_set('UTC');
  * Feel free to switch this to MySQL, it will definitely be better for higher
  * concurrency.
  */
-$pdo = new \PDO('sqlite:data/db.sqlite');
+$pdo = new PDO('sqlite:data/db.sqlite');
 $pdo->setAttribute(PDO::ATTR_ERRMODE, PDO::ERRMODE_EXCEPTION);
 
 // Autoloader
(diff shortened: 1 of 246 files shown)


ChristophWurst (approved):

phil-davis:
Looking at the `phpstan` fails. I will push a small commit...

lib/DAV/TemporaryFileFilterPlugin.php
```diff
@@ -103,8 +103,6 @@ public function initialize(Server $server)
      *
      * This method intercepts any GET, DELETE, PUT and PROPFIND calls to
      * filenames that are known to match the 'temporary file' regex.
-     *
-     * @return bool
```
phil-davis, line 107: This method does not return anything in any other implementation, and has no return type declaration.
I don't know why the PHPdoc here has `@return bool`

So I removed it to make `phpstan` happy.

lib/DAV/TemporaryFileFilterPlugin.php
```diff
@@ -134,17 +132,13 @@ public function beforeMethod(RequestInterface $request, ResponseInterface $respo
      * @param resource $data
      * @param bool     $modified should be set to true, if this event handler
      *                           changed &$data
-     *
-     * @return bool
```
phil-davis, line 140: This method does not return anything in any other implementation, and has no return type declaration.
I don't know why the PHPdoc here has `@return bool`

So I removed it to make `phpstan` happy.

examples/groupwareserver.php
```diff
@@ -63,39 +63,39 @@
...
+$server->addPlugin(new Sabre\CalDAV\Plugin());
+$server->addPlugin(new Sabre\CalDAV\Schedule\Plugin());
+$server->addPlugin(new Sabre\CalDAV\SharingPlugin());
+$server->addPlugin(new Sabre\CalDAV\ICSExportPlugin());
 
 // CardDAV plugins
-$server->addPlugin(new \Sabre\CardDAV\Plugin());
-$server->addPlugin(new \Sabre\CardDAV\VCFExportPlugin());
+$server->addPlugin(new Sabre\CardDAV\Plugin());
+$server->addPlugin(new Sabre\CardDAV\VCFExportPlugin());
```
phil-davis, line 98: I suppose that cs-fixer removed the leading `\` on these files in build and examples because they are not themselves in a namespace. So `\Sabre\Something\...` is the same thing as `Sabre\Something\...`

phil-davis:
There are 246 files changed, and more than 1k lines. I do feel like I should scroll through this to see if there is anything unusual done. Having food now, let's see if I can review it in about 2 hours.

lib/CalDAV/SharingPlugin.php
```diff
@@ -224,8 +224,8 @@ public function httpPost(RequestInterface $request, ResponseInterface $response)
                 // Breaking the event chain
                 return false;
 
-            // The invite-reply document is sent when the user replies to an
-            // invitation of a calendar share.
+                // The invite-reply document is sent when the user replies to an
+                // invitation of a calendar share.
```
phil-davis, line 228: This sort of change is unfortunate. The comment is actually about the following case, but php-cs-fixer wants to indent it the same as the code above (that has already `return`ed.

It does not break anything, so just put up with it for now.

phil-davis (approved):

PR #1664, opened 2026-07-14
refactor: Add more types to the DAV plugins

CarlSchwan:


lib/CalDAV/Plugin.php
phil-davis, line 112: This can return nothing, null, void - sadly.

I suppose soeone should look at what really should happen at this point - throw an exception, return an empty string or what?

CarlSchwan: Reverted that specific change :( 

Looking at the getCalendarHomeForPrincipal usage, returning void is likely not something the users expects. I'll create a seperate PR that throw an InvalidArgument exception

DerDreschner (approved):

lib/CalDAV/Schedule/Plugin.php
```diff
@@ -824,9 +808,9 @@ protected function handleFreeBusyRequest(IOutbox $outbox, VObject\Component $vOb
      *
      * @param string $email address
      *
-     * @return array
+     * @return array{calendar-data: VObject\Component, request-status: string, href: string}
```
phil-davis: If something goes wrong, then request-status and href are returned, but no calendar-data.
Maybe calendar-data can be documented as "optional".

lib/CalDAV/Plugin.php
```diff
@@ -238,11 +226,11 @@ public function initialize(DAV\Server $server)
     /**
      * This functions handles REPORT requests specific to CalDAV.
      *
-     * @param string $reportName
+     * @paran Xml\Request\CalendarQueryReport|CalendarMultiGetReport|Xml\Request\FreeBusyQueryReport $report
```
phil-davis:
```suggestion
     * @param Xml\Request\CalendarQueryReport|CalendarMultiGetReport|Xml\Request\FreeBusyQueryReport $report
```

lib/CardDAV/Plugin.php
```diff
@@ -175,12 +169,11 @@ public function propFindEarly(DAV\PropFind $propFind, DAV\INode $node)
     /**
      * This functions handles REPORT requests specific to CardDAV.
      *
-     * @param string   $reportName
-     * @param \DOMNode $dom
+     * @psalm Xml\Request\AddressBookMultiGetReport $dom
```
phil-davis:
```suggestion
     * @param Xml\Request\AddressBookMultiGetReport $dom
```

Why does this one have `psalm` ?

lib/DAV/Locks/Plugin.php
```diff
@@ -114,12 +98,10 @@ public function getHTTPMethods($uri)
...
      * In this case this is only the number 2. The 2 in the Dav: header
      * indicates the server supports locks.
-     *
-     * @return array
      */
-    public function getFeatures()
+    public function getFeatures(): array
     {
-        return [2];
+        return ['2'];
```
phil-davis, line 104: Probably "a good thing", the '2' actually comes as a string, and easiest to always process and compare it ias a string.

lib/DAV/Xml/Property/Href.php
```diff
@@ -64,18 +62,16 @@ public function __construct($hrefs)
...
         return $this->hrefs[0] ?? null;
     }
 
     /**
      * Returns the hrefs as an array.
      *
-     * @return array
+     * @return list<string>
      */
     public function getHrefs()
```
phil-davis: Does this always return `array` ?

lib/DAV/ServerPlugin.php
```diff
@@ -31,7 +31,7 @@ abstract public function initialize(Server $server);
      * This is for example 'versioning' and is added to the DAV: header
      * in an OPTIONS response.
      *
-     * @return array
+     * @return list<non-empty-string>
      */
     public function getFeatures()
```
phil-davis:
```suggestion
    public function getFeatures(): array
```

KI guess that this has to always return an `array`

lib/DAV/ServerPlugin.php
```diff
@@ -45,11 +45,9 @@ public function getFeatures()
      * This method is passed a uri. It should only return HTTP methods that are
      * available for the specified uri.
      *
-     * @param string $path
-     *
-     * @return array
+     * @return list<non-empty-string>
      */
-    public function getHTTPMethods($path)
+    public function getHTTPMethods(string $path)
```
phil-davis:
```suggestion
    public function getHTTPMethods(string $path): array
```

lib/DAV/ServerPlugin.php
```diff
@@ -60,7 +58,7 @@ public function getHTTPMethods($path)
      * Using this name other plugins will be able to access other plugins
      * using \Sabre\DAV\Server::getPlugin
      *
-     * @return string
+     * @return non-empty-string
      */
     public function getPluginName()
```
phil-davis:
```suggestion
    public function getPluginName(): string
```

lib/DAV/ServerPlugin.php
```diff
@@ -74,11 +72,9 @@ public function getPluginName()
      * Note that you still need to subscribe to the 'report' event to actually
      * implement them
      *
-     * @param string $uri
-     *
      * @return array
      */
-    public function getSupportedReportSet($uri)
+    public function getSupportedReportSet(string $uri)
```
phil-davis:
```suggestion
    public function getSupportedReportSet(string $uri): array
```

lib/DAV/ServerPlugin.php
```diff
@@ -92,7 +88,7 @@ public function getSupportedReportSet($uri)
      * The description key in the returned array may contain html and will not
      * be sanitized.
      *
-     * @return array
+     * @return array{name: non-empty-string, description: ?string, link: ?string}
      */
     public function getPluginInfo()
```
phil-davis:
```suggestion
    public function getPluginInfo(): array
```

lib/DAVACL/Plugin.php
```diff
@@ -113,20 +113,16 @@ class Plugin extends DAV\ServerPlugin
      * Returns a list of features added by this plugin.
      *
      * This list is used in the response of a HTTP OPTIONS request.
-     *
-     * @return list<string>
```
phil-davis, line 117: Keep the PHPdoc, it is more specific than just `array`

lib/DAVACL/Plugin.php
```diff
@@ -113,20 +113,16 @@ class Plugin extends DAV\ServerPlugin
...
-    public function getFeatures()
+    public function getFeatures(): array
     {
         return ['access-control', 'calendarserver-principal-property-search'];
     }
 
     /**
      * Returns a list of available methods for a given url.
-     *
-     * @return list<string>
```
phil-davis, line 127: Keep the PHPdoc, it is more specific than just `array`

phil-davis:
I made a few adjustments here, and added a few more actual `array` declarations where something really does always return an array.

The next release from `master` is planned to be a major release, supporting PHP 7.4 and up. So we can add parameter and return type declarations where we are confident in what they have to be.

phil-davis (approved):

Issue #1601, opened 2025-10-26
free-busy-request not working


wlamers:
I have placed this on the Baikal github as well, but I am now not so sure anymore if it is Baikal related. Hope someone can shed some light on this?

It might be related to https://github.com/sabre-io/dav/issues/1185
That's a very old thread, so not so likely that nobody else bumped into this in the mean time.

Really hope someone has a clue!

Baikal version
0.10.1

PHP version
8.1.33

Steps to reproduce
Requesting free-busy report from other users on the server, using several CalDAV clients like Thunderbird and Evolution.

Expected behaviour
Free-busy information being reported.

Current behaviour
There is no result shown (empty).

I have done many hours of debugging with Wireshark. Also changed from Sqlite to Mysql and from digest to basic auth.

It seems that Thunderbird is sending these request to url: '/dav.php/calendars/test/outbox/'
I then also have imitated the request using Python scripts and might have found a clue (No calendar-home-set property found). See the logs. I have tried to add "calendar-home-set" to the mysql database, but to no avail.

### Logs

```
Response body: <?xml version="1.0" encoding="utf-8"?>
<cal:schedule-response xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns" xmlns:cal="urn:ietf:params:xml:ns:caldav" xmlns:cs="http://calendarserver.org/ns/" xmlns:card="urn:ietf:params:xml:ns:carddav">
  <cal:response>
    <cal:recipient>
      <d:href>mailto:redact@redact.com</d:href>
    </cal:recipient>
    <cal:request-status>3.7;No calendar-home-set property found</cal:request-status>
  </cal:response>
</cal:schedule-response>
```

What seems to work: I can retrieve the free-busy info if I send a request (using the below Python script) to the user calendar path directly. E.g. to /dav.php/calendars/test/thecalendarname/

```
import requests
from requests.auth import HTTPDigestAuth
import xml.etree.ElementTree as ET
from datetime import datetime, timedelta

USERNAME = "user@redacted.com"
PASSWORD = "pwd"
CALDAV_URL = "http://server/dav.php/calendars/test/thecalendarname/"

START = "20251023T000000Z"
END = "20251024T000000Z"

# CalDAV free-busy REPORT XML body
FREE_BUSY_REPORT = f"""<?xml version="1.0" encoding="UTF-8"?>
<c:free-busy-query xmlns:c="urn:ietf:params:xml:ns:caldav">
[...]

PR #1673, opened 2026-09-11
fix(caldav): allow free-busy requests for other users' principals

### LINKED ISSUE #1185: FreeBusy-Report fails with Status 3.7;No calendar-home-set property found
This happens with Sabre 3.2.2. When inviting people with thunderbird, FreeBusy-information is displayed only for some attendees.

In order to produce a FreeBusy-report principalSearch() from DAVACL/Plugin.php is called to find the principal by email-address and get some properties. 

```
    function principalSearch(array $searchProperties, array $requestedProperties, $collectionUri = null, $test = 'allof') {
        if (!is_null($collectionUri)) {
            $uris = [$collectionUri];
        } else {
            $uris = $this->principalCollectionSet;
        }
        $lookupResults = [];
        foreach ($uris as $uri) {
            $principalCollection = $this->server->tree->getNodeForPath($uri);
            if (!$principalCollection instanceof IPrincipalCollection) {
                // Not a principal collection, we're simply going to ignore
                // this.
                continue;
            }
            $results = $principalCollection->searchPrincipals($searchProperties, $test);
            foreach ($results as $result) {
                $lookupResults[] = rtrim($uri, '/') . '/' . $result;
            }
        }
        $matches = [];
        foreach ($lookupResults as $lookupResult) {
            list($matches[]) = $this->server->getPropertiesForPath($lookupResult, $requestedProperties, 0);
        }
```
$result will contain one principal in all cases, but getPropertiesForPath() does not find the calendar-home-set property for some principals.

Any ideas why this is happening for some principals only?

The calendar-home-set property of principals/xxx should be $baseUri/calendars/xxx. It's not stored anywhere but will be calculated for every principal. So why should this calculation fail for some principals? So what am I missing here?

Peter

thilo:
Fixes #1185 (also reported as #1601 and sabre-io/Baikal#1360).

## Problem

A free-busy request (POST to the outbox) for another user on the same server
answers `3.7;No calendar-home-set property found` unless the requester is an
admin. `getFreeBusyForEmail()` finds the recipient with `principalSearch()`,
which reads `calendar-home-set` and `schedule-inbox-URL` through the ACL. A
principal is readable only by itself (`{DAV:}all` for `{DAV:}owner`), so for
any other user the properties come back 403. With `hideNodesFromListings` the
lookup also triggers an `Undefined array key 0` warning in `principalSearch()`.

The existing tests pass because they make the requesting user an admin first.

## Change

Run the lookup with the ACL `propFind` listener removed, the same way
`scheduleLocalDelivery()` already looks up the recipient's inbox, and put it
back in a `finally`.

## Privacy

This was proposed in #1185 before, and @evert
[declined it](https://github.com/sabre-io/dav/issues/1185#issuecomment-628793168)
for privacy reasons: the default should not let everyone read everyone's
information. I think the change keeps that default, because the principal ACL
is not what decides free-busy access:

- The principal properties are only used to find the recipient's inbox and
  calendars. They are not sent to the client.
- Whether free-busy may be queried is decided by the `schedule-query-freebusy`
  check on the recipient's inbox, which comes right after the lookup. RFC 6638
  defines that privilege on the scheduling inbox for this purpose. If the check
  fails, `checkPrivileges()` throws `NeedPrivileges` and the request fails with
  403
  ([raised by @m-a-v](https://github.com/sabre-io/dav/issues/1185#issuecomment-628805616),
  now covered by a test).
- sabre's `Schedule\Inbox::getACL()` already grants `schedule-deliver`, which
  includes `schedule-query-freebusy`, to `{DAV:}authenticated`. The inbox
  already allows the query, and the principal ACL blocked it before that
  check was reached.
- The response contains only busy periods, not event details.
- It reveals nothing new about which addresses have accounts. Master already
  answers `3.7;Could not find principal` for unknown addresses and
  `3.7;No calendar-home-set property found` for known ones.

This is a behaviour change: on a default setup, authenticated users can now
query each other's free-busy, as the inbox ACL says they may. To keep
free-busy private, restrict the inbox instead of the principal: grant
`schedule-deliver-invite` and `schedule-deliver-reply` rather than
`schedule-deliver`. Invitations keep working and free-busy queries are refused.
`testFailWithoutFreeBusyPrivilege` does exactly that.

If you'd rather keep the old behaviour as the default, I can put this behind
an opt-in property on the Schedule plugin instead.

## Tests

- `testSucceedWithoutAdmin` and `testSucceedWithoutAdminWithHiddenNodes` in
  `FreeBusyRequestTest` repeat `testSucceed` without the admin shortcut, once
  with and once without `hideNodesFromListings`. Both fail on master and pass
  with the change. They also check that the ACL is active again afterwards.
- `testFailWithoutFreeBusyPrivilege` restricts the recipient's inbox as
  described above and expects `NeedPrivileges`. It fails if the
  `schedule-query-freebusy` check is removed.
- `tests/Sabre/CalDAV/Schedule` passes except
  `SchedulingObjectTest::testGetLastModified`, which errors on master too
  (PHP 8.3 `ctype_digit()` deprecation).

Found running Davis (sabre/dav 4.7.1) with macOS Calendar, which shows "There
was a problem accessing availability" when an event has attendees on the same
server.


thilo:
A downstream data point: Baikal is working around the same bug on its own side in sabre-io/Baikal#1406. It grants the organizer temporary `calendar-proxy-read` membership on each attendee principal for the duration of the outbox POST, behind an `allow_free_busy_lookup` option that defaults to on.

That is the same conclusion this PR reaches, that free-busy between users of one server should work without extra configuration, but it has to be re-implemented in each downstream project (it subclasses both the principal backend and the scheduling plugin). If the default here is deliberate, a note in the docs on where to hook in would probably save that work; if not, the inbox privilege already expresses it.

